PR #3, opened 2023-04-26
:bug: Fix base url for the new repo

Itxaka:


Itxaka:
Editing pages and linking to their source now works! Good catch @mudler !

mauromorales (approved):
lgtm

PR #7, opened 2023-05-12
doc: Adding references to EdgeVPN network token format in getting started...

nicop311:
... and Installation section

At my first view of the Kairos **getting started** section, I was confused by what the QR Code was. And I was confused by the term `TOKEN` or `NETWORK_TOKEN` which was use but not necessary explained in the getting started or Installation section. Therefore I propose to add a few explanation on the EdgeVPN token in those sections which are the first sections that a reader reads.


Itxaka (approved):
very nice, thank you!

PR #11, opened 2023-05-18
Correct connman config for static ip and nameserver

mauromorales:
Fixes https://github.com/kairos-io/kairos/issues/335

### DIFF
--- a/content/en/docs/Advanced/networking.md
+++ b/content/en/docs/Advanced/networking.md
@@ -27,8 +27,9 @@ stages:
           content: |
             [service_eth0]
             Type = ethernet
-            IPv4 = 10.1.1.1/16/10.1.0.1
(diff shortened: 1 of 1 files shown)


Itxaka (approved):
:+1: I guess ? :D

content/en/docs/Advanced/networking.md
```diff
@@ -27,8 +27,9 @@ stages:
           content: |
             [service_eth0]
             Type = ethernet
-            IPv4 = 10.1.1.1/16/10.1.0.1
```
jimmykarily, line 30: These IP addresses (and the one below) seem to match edgevpn network IPs. Maybe we should have an edgevpn specific example too. Regardless of the IPs, it was the wrong syntax apparently, so :+1: .

PR #12, opened 2023-05-19
Update docs to release 2.1.0

mauromorales:


Itxaka (approved):
oh this is bad, we should have this as part of the release or something.

I wonder if we can trick renovate to do this for us :D

mauromorales:
yeah indeed, we need to automate it :P

PR #15, opened 2023-05-23
:robot: Use renovate to update kairos version

mauromorales:


Itxaka (approved):
oh wow, this is possible? Pretty cool!

Wondering if we should also use renovate directly on our packages instead of our custom bumper...

mauromorales:
@Itxaka good point, as long as they are accessible via a github release tag or a docker container version, it should be possible

PR #23, opened 2023-05-23
Add warning to RPi config example for LVM flag

mauromorales:


content/en/docs/Installation/raspberry.md
```diff
@@ -55,6 +55,11 @@ You can push additional `cloud config` files. For a full reference check out the
 
 The following shell script shows how to locally rebuild and customize the image with docker
 
+
+{{% alert title="Warning" %}}
+If you're using osbuilder between versions 0.6.0 and 0.6.5, you need to pass the flag `--use-lvm` to the `build-arm-image.sh` script, the same way you pass `--local`.
```
Itxaka, line 60: to be fair, I would leave the example as is (wont affect 0.6.6 to have that flag, it just does nothing) and will instead fix this to say that the lvm flag is deprecated and from 0.6.6 onwards it does nothing.

Then down the line we can remove the lvm flag fromt he docs directly.

What do you think?

mauromorales: yup makes sense, updated it

Itxaka (approved):

PR #41, opened 2023-06-26
Add custom partitioning section to documentation

mauromorales:
Fixes https://github.com/kairos-io/kairos/issues/209

### DIFF
--- a/content/en/docs/Reference/configuration.md
+++ b/content/en/docs/Reference/configuration.md
@@ -1043,6 +1043,11 @@ You're responsible to make sure the sizes of the partitions fit properly within 
 the agent, but they will not fail the installation process unless you pass the `--strict` flag.
 {{% /alert %}}
 
+{{% alert title="Warning" %}}
+In the case of multiple devices, make sure you don't choose `auto` to determine on which device to install but instead to point
+the installation to the device where you are creating the custom partitions.
+{{% /alert %}}
+
 ```shell
 #cloud-config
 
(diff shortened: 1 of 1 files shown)


content/en/docs/Reference/configuration.md
```diff
@@ -1034,6 +1034,51 @@ stages:
...
+{{% alert title="Warning" %}}
+You're responsible to make sure the sizes of the partitions fit properly within the disk. Issues of space will be highlighted by
+the agent, but they will not fail the installation process unless you pass the `--strict` flag.
+{{% /alert %}}
+
+```shell
+#cloud-config
+
+install:
+  no-format: true
```
jimmykarily, line 1050: Let's add some comments to explain what each section does

```suggestion
  # Make sure the installer won't delete our custom partitions
  no-format: true
```

content/en/docs/Reference/configuration.md
```diff
@@ -1034,6 +1034,51 @@ stages:
...
+
+stages:
+  kairos-install.pre.before:
+  # TODO: check which device exists
+  - if:  '[ -e /dev/vda ] && (kairos-agent state get boot | grep -q livecd_boot)'
```
jimmykarily, line 1055: Originally I added this check because I used the `boot` stage. I wanted to avoid partitioning the disk every time the system booted. Since then, I changed to this `kairos-install.pre.before` stage which I think is safe to run without a condition. We can keep the check for the disk though. E.g. maybe the user uses the same config on 2 different systems where the disk may be named differently.

content/en/docs/Reference/configuration.md
```diff
@@ -1034,6 +1034,51 @@ stages:
...
+
+```shell
+#cloud-config
+
+install:
+  no-format: true
+
+stages:
+  kairos-install.pre.before:
+  # TODO: check which device exists
```
jimmykarily, line 1054: Let's change this to something that explains the condition below.

content/en/docs/Reference/configuration.md
mauromorales: ok will adapt

jimmykarily (approved):

PR #47, opened 2023-07-05
Move links to be relative to the current dir

Itxaka:
Instead of pointing to the base_url + /docs/ as default. This should make the documentation work anywhere, no matter if we change the dir in which the docs are contained as they are referencing by relative paths

This also has the advantage that hugo will check the validity of the links on compilation and fail if a link points to nowhere

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,6 +3,12 @@
   "extends": [
     "config:base"
   ],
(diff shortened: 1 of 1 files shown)


mauromorales (approved):
fantastic 😅

PR #48, opened 2023-07-05
Move links to be relative to the current dir

Itxaka:
Instead of pointing to the base_url + /docs/ as default. This should make the documentation work anywhere, no matter if we change the dir in which the docs are contained as they are referencing by relative paths

This also has the advantage that hugo will check the validity of the links on compilation and fail if a link points to nowhere

### DIFF
--- a/content/en/docs/Advanced/build.md
+++ b/content/en/docs/Advanced/build.md
@@ -21,7 +21,7 @@ Note the same can be achieved by using advanced configuration and actually modif
 
 To achieve that, Kairos provides a set of Kubernetes Native Extensions that allow to programmatically generate Installable mediums, Cloud Images and Netboot artifacts. These provide on-demand customization and exploit Kubernetes patterns to automatically provision nodes using control-plane management clusters - however, the same toolset can be used to build appliances for local development and debugging.
 
-The [automated](/docs/installation/automated) section already shows some examples of how to leverage the Kubernetes Native Extensions and use the Kairos images to build appliances, in this section we will cover and describe in detail how to leverage the CRDs and the Kairos factory to build custom appliances.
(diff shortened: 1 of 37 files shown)


content/en/docs/Advanced/build.md
```diff
@@ -21,7 +21,7 @@ Note the same can be achieved by using advanced configuration and actually modif
 
 To achieve that, Kairos provides a set of Kubernetes Native Extensions that allow to programmatically generate Installable mediums, Cloud Images and Netboot artifacts. These provide on-demand customization and exploit Kubernetes patterns to automatically provision nodes using control-plane management clusters - however, the same toolset can be used to build appliances for local development and debugging.
 
-The [automated](/docs/installation/automated) section already shows some examples of how to leverage the Kubernetes Native Extensions and use the Kairos images to build appliances, in this section we will cover and describe in detail how to leverage the CRDs and the Kairos factory to build custom appliances.
+The [automated]({{< ref "../installation/automated" >}}) section already shows some examples of how to leverage the Kubernetes Native Extensions and use the Kairos images to build appliances, in this section we will cover and describe in detail how to leverage the CRDs and the Kairos factory to build custom appliances.
```
mauromorales, line 24: Just noticed that this build a URL with the domain specified on the configuration, which means that it breaks the previews. The best would be to use `relref` https://gohugo.io/functions/relref/

e.g. https://deploy-preview-48--kairos-io.netlify.app/docs/advanced/build/ points to https://kairos.io/docs/installation/automated/

Itxaka: :sob:

mauromorales (approved):
sweeeet!

PR #54, opened 2023-07-18
change kairos-cli to kairosctl

si458:
Signed-off-by: si458 <simonsmith5521@gmail.com>


mauromorales (approved):
thanks!

si458:
> thanks!

No worries, spent a whole hour trying to download the old file then realised it was called something else

PR #56, opened 2023-07-20
Fix older elemental-cli reference

Itxaka:
Fixes https://github.com/kairos-io/kairos/issues/1623


mauromorales (approved):
lgtm ... I wish we had versioned documentation cuz I guess this won't work with some of the older releases

PR #59, opened 2023-08-02
Update cloud config and add partitions config

Itxaka:
Updates the cloud config page with all the now possible values to override th default configs, now that ythe full config and spec is read from the cloud-config via the agent.

Also adds a page on how to specifically configure the current partitions during install and add extra partitions during install

Fixes: https://github.com/kairos-io/kairos/issues/558

### DIFF
--- a/content/en/docs/Advanced/configuring_partitions.md
+++ b/content/en/docs/Advanced/configuring_partitions.md
@@ -1,0 +1,95 @@
+---
+title: "Configuring partitions"
+linkTitle: "Configuring partitions"
+weight: 1
+description: >
+---
+
+{{% alert title="Note" color="warning" %}}
+
+This feature will be available in Kairos version `2.4.0` and in all future releases.
+
+{{% /alert %}}
+
+
+Kairos configuration mechanism for partitions is based on the [cloud-config]({{< relref "../Reference/configuration" >}}) file 
+given during installation to override the default values set by the installer.
+
+We allow certain flexibility in the sizes and filesystems used for the default install and allow to create extra partitions as well.
+
+For example, the following cloud-config will make the `oem` partition have a size of 512Mb and an `ext4` filesystem, 
+recovery with a size of 10000Mb and a `ext4` filesystem, while leaving the rest of the partitions to their default sizes and filesystems. 
+
+```yaml
+#cloud-config
+
+install:
+  device: "/dev/sda"
+  auto: true
+  partitions:
+    oem:
+      size: 512
+      fs: ext4
+    recovery:
+      size: 10000
+      fs: ext4
+```
+
+
+
+And the following config will leave the default partitions as is, but create 2 new extra partitions with the given sizes, filesystems and labels:
+
+```yaml
+#cloud-config
+
+install:
+  device: "/dev/sda"
+  auto: true
+  extra-partitions:
+    - name: first_partition
+      size: 512
+      fs: ext3
+    - name: second_partition
+      size: 100
+      fs: ext2
+      label: PARTITION_TWO
+```
+
+In the case of wanting an extra partition to take over the rest of the disk space (by setting the size to 0), we need to also mark persistent with a fixed size.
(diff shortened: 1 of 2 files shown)


content/en/docs/Advanced/configuring_partitions.md
```diff
@@ -0,0 +1,95 @@
...
+      fs: ext3
+    - name: second_partition
+      size: 100
+      fs: ext2
+      label: PARTITION_TWO
+```
+
+In the case of wanting an extra partition to take over the rest of the disk space (by setting the size to 0), we need to also mark persistent with a fixed size.
+Currently, the default value for the persistent partition is set to 0 which means "take over the rest of the disk space free".
+In order to mark a different partition with that value, we need to set it to a fixed size, as there cannot be 2 partitions that want to take over the rest of the space.
```
mauromorales, line 60: I think this sounds a bit more complex than it needs to be. Maybe something like: only one partition can expand to the rest of the disk. Either persistent or one of the extra-partitions. In case you want the latter, you need to specify the size of persistent.

content/en/docs/Advanced/configuring_partitions.md
```diff
@@ -0,0 +1,95 @@
...
+    - name: big_partition
+      size: 0
+      fs: ext3
+```
+
+
+Note that there are some caveats in the `extra partitions` setup:
+ - only `size`, `fs`, `name` and `label` are used for the partition creation, the name is currently used for the partition label.
+ - if a partition has no fs set, the partition will be created, but it will not be formatted
+ - No mounting of any type is done during installation to the extra partitions. That part should be done on the stages of the cloud-config manually, with something like the following step:
```
mauromorales, line 83: start with capital while the others without

content/en/docs/Advanced/configuring_partitions.md
```diff
@@ -0,0 +1,95 @@
...
+  extra-partitions:
+    - name: big_partition
+      size: 0
+      fs: ext3
+```
+
+
+Note that there are some caveats in the `extra partitions` setup:
+ - only `size`, `fs`, `name` and `label` are used for the partition creation, the name is currently used for the partition label.
+ - if a partition has no fs set, the partition will be created, but it will not be formatted
```
mauromorales, line 82: ends without punctuation while the others do

mauromorales (approved):
LGTM

content/en/docs/Reference/configuration.md
```diff
@@ -22,6 +22,14 @@ users:
   ssh_authorized_keys:
   # - github:mudler
 
+# enable debug logging
+debug: true
```
jimmykarily, line 26: After the merging of all configs, does the schema need an update? https://github.com/kairos-io/kairos-sdk/tree/main/schema

Itxaka: Good point! I remember updating it already so it would pass the tests. Not sure if the tests are now broken or it's not necessary anymore...

Gonna check

content/en/docs/Advanced/configuring_partitions.md
```diff
@@ -0,0 +1,95 @@
...
+    persistent:
+      size: 500
+  extra-partitions:
+    - name: big_partition
+      size: 0
+      fs: ext3
+```
+
+
+Note that there are some caveats in the `extra partitions` setup:
```
jimmykarily, line 80: Shouldn't we also mention that if they try to override the labels of the default partitions, these will be ignored? (as per the PR here: https://github.com/kairos-io/kairos-agent/pull/108/files)

Itxaka: Not really no, the label stuff is for the system partitions and its mentioned above that we only support fs and size overrides.

Itxaka: I dont want to give them ideas that they can indeed override the labels lol, security through obscurity

PR #62, opened 2023-08-03
Improve sidebar

Itxaka:
Adds icons to each entry
Nav bar is collapsed by default as it now covers the full side on any monitor and hides part of the docs

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -128,6 +128,8 @@ registryURL = "quay.io/kairos"
 [params.ui]
 # Enable to show the side bar menu in its compact state.
 sidebar_menu_compact = true
+# Enable to show the foldable triangle
+sidebar_menu_foldable = true
 #  Set to true to disable breadcrumb navigation.
 breadcrumb_disable = false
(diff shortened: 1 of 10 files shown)


Itxaka:
image

Itxaka:
Could also add the foldable to indicate that the sections can be expanded:

image

mauromorales (approved):
😎

Itxaka:
> @Itxaka does the foldable take care that the open one has the triangle looking down? otherwise better not add

yep, its theusual triangle thing that points down when expanded, and rigth when not.

Note that that is not part of this PR as I was not sure of adding it. IMHO it looks uglier but its a good UX to show that it can be expanded...

Itxaka:
cc @jimmykarily @mudler any thougths on this? foldable vs non foldable, nice looking vs good UX, icons opinions?

jimmykarily:
I like the arrows. Icons look good too.

Itxaka:
arrows it is then.

PR #65, opened 2023-08-07
Homepage-revamp

mirceabozi:


mauromorales:
closing because the cache of netlify is braking, will follow up in #70

PR #84, opened 2023-09-14
Link to wifi example and add shortcode to include remote source

Itxaka:
This allows us to include the examples directly from the kairos repo if needed


jimmykarily (approved):
Nice!

PR #90, opened 2023-09-25
Update versions to latest released

mauromorales:
fixes #1851


mudler (commented):
wondering why renovate didn't pick this up?

mudler (approved):

mauromorales:
@mudler good call! it didn't catch it because the name of the config file changed from config.toml to hugo.toml in the latest version of hugo, and our renovate.json was pointing to the latter. Adding the fix to this PR

PR #102, opened 2023-10-30
Reference new building args

mauromorales:
:warning: wait for the release before merging

Fixes kairos-io/kairos#1964


content/en/docs/Development/development.md
```diff
@@ -13,27 +13,41 @@ Here you can find development notes intended for maintainers and guidance for ne
...
+  --FLAVOR_RELEASE=38 \
+  --BASE_IMAGE=fedora:38 \
+  --MODEL=generic \
+  --VARIANT=core
 ```
 
+- **Flavor**: is the distribution (the name flavor is kept for backwards compatibility). In this case `fedora`
+- **Family**: is a group of distributions that get built in a very similar way. In this case `rhel`, which can build `fedora`, `rockylinux` and `almalinux`.
+- **Flavor release**: is the version of the distribution. In this case `38`
+- **Base image**: is the base image used to build the distribution. In this case `fedora**:38`, but you could use your own image which itself is based on `fedora**:38`.
```
jimmykarily, line 40: what's with the starts in `fedora**:38` ?

jimmykarily (approved):

content/en/docs/Development/development.md
mauromorales: LOL i went too far with my s/:/:**/g ... good catch thanks

PR #103, opened 2023-10-31
Fix artifact names

jimmykarily:
Part of: https://github.com/kairos-io/kairos/issues/1978

Fixes kairos-io/kairos#103


content/en/docs/Reference/image_matrix.md
```diff
@@ -16,81 +16,35 @@ These images are pushed to quay.io and are available for installation and upgrad
 
 Kairos release processes generates images based on official container images from popular Linux distributions. If you don't see your preferred distribution, check if [we are already planning](https://github.com/kairos-io/kairos/issues?q=is%3Aopen+is%3Aissue+label%3Aarea%2Fflavor) support for it or create a new issue.
 
-Below is a list of the available images and their locations on the quay.io registry:
+Below is a list of the container repositories for each flavor:
```
mauromorales, line 19: This change makes it a lot clearer but I would still at least add a link to the latest release artifacts, so they can also download ISOs or other artifacts

hugo.toml
```diff
@@ -125,6 +125,9 @@ auroraboot_version = "v0.2.6"
 flavor = "debian"
 armFlavor = "alpine"
 registryURL = "quay.io/kairos"
+exampleImage = "opensuse"
+exampleImageCoreTag = "leap-15.5-core-amd64-generic-v2.4.1"
+exampleImageStandardTag = "leap-15.5-core-amd64-generic-v2.4.1-k3sv1.27.3-k3s1"
```
mauromorales, line 130:
```suggestion
exampleImageCoreTag = "leap-15.5-core-amd64-generic-v2.4.1"
exampleImageStandardTag = "leap-15.5-standard-amd64-generic-v2.4.1-k3sv1.27.3-k3s1"
```

Standard one is incorrect here, but overall, I would prefer not to do this here since it will keep the versions of the examples pinned. I think it would be better to build them using the mechanism we already have with the version. 

```
"{{< registryURL >}}/{{< exampleImage >}}:leap-15.5-core-amd64-generic-{{< kairosVersion >}}"
```

Or even better, maybe introduce another function which itself does the code above and can be used like so:

```
"{{< registryURL >}}/{{< exampleStandardImage >}}
```

maybe even go as far as

```
{{< exampleStandardImage >}}
```

PR #104, opened 2023-11-13
Mention rd immucore debug

jimmykarily:
https://github.com/kairos-io/kairos/issues/1989


Itxaka (approved):

Itxaka:
Maybe we should link to the immucore readme somewhere as it documents several options that can be toggled via cmdline

mauromorales (approved):

jimmykarily:
> Maybe we should link to the immucore readme somewhere as it documents several options that can be toggled via cmdline

Good point. Added.

PR #108, opened 2023-11-16
1962 document k3s args

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/1962

Currently blocked by this: https://github.com/kairos-io/kairos/issues/2012

### DIFF
--- a/content/en/docs/Advanced/bundles.md
+++ b/content/en/docs/Advanced/bundles.md
@@ -67,6 +67,12 @@ metallb:
   address_pool: 192.168.1.10-192.168.1.20
 ```
 
+{{% alert title="Warning" %}}
+When a bundle is installed after Kairos installation (see above), changing the
+bundle's configuration block after Kairos installation is complete,
+will not have any effect.
+{{% /alert %}}
+
 ## Bundle types
 
 Bundles can carry also binaries that can be overlayed in the rootfs, either while [building images]({{< relref "build" >}}) or with [Live layering]({{< relref "livelayering" >}}).
--- a/content/en/docs/Advanced/bundles.md
+++ b/content/en/docs/Advanced/bundles.md
@@ -90,6 +90,13 @@ the installation only happens once. Changing the bundle's configuration block
 after Kairos installation is complete, will not have any effect.
 {{% /alert %}}
 
+If you want the installation to stop if a bundle installation fails, you can
+add set the following option to `true` in your Kairos config:
+
+```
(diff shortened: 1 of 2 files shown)


content/en/docs/Advanced/bundles.md
```diff
@@ -67,6 +67,12 @@ metallb:
   address_pool: 192.168.1.10-192.168.1.20
 ```
 
+When a bundle is installed after Kairos installation (see above), changing the
+bundle's configuration block after Kairos installation is complete,
+will not have any effect.
+{{% /alert %}}
```
jimmykarily, line 74: I'm not sure about this one. We need to try this out with both types of bundle installation: https://kairos.io/docs/advanced/bundles/#consuming-bundles and verify. E.g. Can I change the configuration for a bundle that gets installed before Kubernetes starts at any time? What does it mean "before Kubernetes starts" when I'm using a core image?

jimmykarily:
I use this config:

```
#cloud-config

k3s:
  enabled: true

fail_on_bundles_errors: true

users:
  - name: kairos
    passwd: kairos

bundles:
- targets:
  - run://jimmykarily/mybundle:latest

mybundle:
  foo: bar
```

My bundle has a run.sh script like this one:

```
#!/bin/bash

echo $(kairos-agent config get mybundle) > /home/kairos/bundle-log
```

After installation and after booting to the installed system, the `/home/kairos/bundle-log` file exists and has the `foo: bar` content.

Conclusion 1: This type of bundle definition in the config, makes the bundle run.sh script run on first boot


Removing the `/home/kairos/bundle-log` file and rebooting doesn't make the file appear again.

Conclusion 2: It on runs on the first boot (as [seen in the code](https://github.com/kairos-io/kairos-agent/blob/944ce8ce741adc24bb4ebbc5a5fba460c1e83e53/internal/agent/hooks/hook.go#L29)) and not on the subsequent boots.

TODO:
- what happens if k3s is not enabled?
- what happens if the images doesn't have k3s at all? (core image)
- try with this type of config to see when it triggers:
  ```
#cloud-config
install:
 bundles:
 - targets:
   - run://<image>
  ```

jimmykarily:
Maybe also answer to this question while at it: https://github.com/kairos-io/kairos/discussions/2016

jimmykarily:
TODO: Document the use of `fail_on_bundles_errors: true`

mauromorales (approved):
much needed explanations, thanks @jimmykarily

PR #119, opened 2023-12-19
Update interactive.md

davidthewatson:
The filename in the present documentation is wrong. I've updated the doc to match the installation.


mauromorales (approved):
@davidthewatson thanks!

PR #130, opened 2024-01-19
Tell the user how to regenerate k3s args

mauromorales:
I'm not sure if this is recommended or not, but I needed to achieve this, and it is the route I took, wdyt?

### DIFF
--- a/content/en/docs/Reference/configuration.md
+++ b/content/en/docs/Reference/configuration.md
@@ -508,7 +508,7 @@ k3s-agent:
 {{< /tabpane >}}
 
 {{% alert title="Warning" %}}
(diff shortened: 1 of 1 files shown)


Itxaka (commented):
Is this really true? Is during the first boot after install, not during install.

Also, if you get a deployed cluster  and then remove the file and restart, will your existing cluster work afterwards? Will it be reinitialized fully?

jimmykarily:
We tried that on a cluster where we didn't care about the workloads on it. It was a hack and I have no idea in what ways it could break things. We should rather not suggest this unless test it thoroughly.

mauromorales:
closing in favor of https://github.com/kairos-io/kairos/issues/2160

PR #132, opened 2024-01-19
Update takeover info

Itxaka:
Add the missing source flag to takeover installs

related: https://github.com/kairos-io/kairos/issues/1528

### DIFF
--- a/content/en/docs/Installation/takeover.md
+++ b/content/en/docs/Installation/takeover.md
@@ -23,7 +23,14 @@ users:
   - github:mudler
 EOF
 export CONFIG_FILE=config.yaml
-docker run --privileged -v $PWD:/data -v /dev:/dev -ti $IMAGE kairos-agent manual-install --device $DEVICE /data/$CONFIG_FILE
+docker run --privileged -v $PWD:/data -v /dev:/dev -ti $IMAGE kairos-agent manual-install --device $DEVICE --source dir:/ /data/$CONFIG_FILE
 ```
-kkgT
+
+{{% alert title="Note" color="warning" %}}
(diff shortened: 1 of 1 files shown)


jimmykarily (approved):

content/en/docs/Installation/takeover.md
```diff
@@ -23,7 +23,14 @@ users:
...
 EOF
 export CONFIG_FILE=config.yaml
-docker run --privileged -v $PWD:/data -v /dev:/dev -ti $IMAGE kairos-agent manual-install --device $DEVICE /data/$CONFIG_FILE
+docker run --privileged -v $PWD:/data -v /dev:/dev -ti $IMAGE kairos-agent manual-install --device $DEVICE --source dir:/ /data/$CONFIG_FILE
 ```
-kkgT
+
+{{% alert title="Note" color="warning" %}}
+
+`--source` flag refers to the source of installation. If you want to install from the pulled docker image you can set the `--source` flag to `dir:/` to use the root dir of the image as install source. Otherwise you can point it to an oci image with the `oci:` prefix.
```
jimmykarily, line 31: Doesn't `kairos-agent` install the current running OS by default? In that case, the `--source` flag shouldn't be needed here, it's already the running image.

Itxaka: for a takeover no, it derfaults to use the isomount dir (/run/rootfsbase) which works on media and pxe installs but thats it. For takeover you need to specify the source exactly as the docker image doesnt have the proper dir (not booted from media so dracut didnt mount the image into the rootfsbase dir)

PR #138, opened 2024-02-01
Update some image names to our new convention

mauromorales:
I also decided to refactor the shortcodes to be displayed so they are easier to read and easier to detect changes

fixes #2192

Signed-off-by: Mauro Morales <mauro.morales@spectrocloud.com>


Itxaka (approved):
Yeah, this is cool, very nice Mauro!

jimmykarily:
Would be nice if we could use versioneer here. But maybe it's not worth the effort (if at all possible).

PR #146, opened 2024-02-08
Add versioning with sub dirs starting with v3

mauromorales:
As soon as v3 is released, we can merge this, and we can make use of versioned documentation. We just need to keep in mind never to remove the vX.Y.Z branches


mudler (approved):
loving it!

jimmykarily:
NOTE: the search is gone (in the deploy preview at least)

mauromorales:
fixed search

jimmykarily (approved):

PR #148, opened 2024-02-19
Trustedboot: update commands to run swtpm separately

mudler:
That allows to boot the EFI files directly off from the disk

### DIFF
--- a/content/en/docs/Installation/trustedboot.md
+++ b/content/en/docs/Installation/trustedboot.md
@@ -124,7 +124,6 @@ WORKDIR /work
 
 RUN <<EOF
 echo "#!/bin/bash -ex" >> /entrypoint.sh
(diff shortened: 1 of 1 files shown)


Itxaka (commented):
This is kind of icky because reusing the same tpm for different keys and such can lead to the tpm blocking and errors coming from it. Maybe we should identify those and add some troubleshooting because it's a pain in the ass and you don't notice that it may be due to that

mudler:
> This is kind of icky because reusing the same tpm for different keys and such can lead to the tpm blocking and errors coming from it. Maybe we should identify those and add some troubleshooting because it's a pain in the ass and you don't notice that it may be due to that

I know, but we need some pointers atm to handle around, and this is so far working quite well here, let's revisit it as we find out better steps to give away as pointers

PR #149, opened 2024-02-21
Add bootentry docs

Itxaka:


Itxaka:
Seems like netlify is broken somehow, other PRs also suffer from the same stuff.

scripts/build.sh
```diff
@@ -21,8 +21,6 @@ rm -rf "${publicpath}" || true
 
 npm install --save-dev autoprefixer postcss-cli postcss
 
-hugo mod get
-hugo mod graph
```
mauromorales, line 25: how come this is removed?

mauromorales: ah I see the comment on the PR

mauromorales: @Itxaka yeah I see, for me this is intentional since I'd like us to find the issues when bumping instead of try to bump after 2 years like we did last time which is a lot more work.

Would you mind keeping them and instead adding this in the netlify.toml

```
[build.environment]
  HUGO_VERSION = "0.113.0"
```

I think that should fix it

Itxaka: Will do!

Itxaka: yep, good one @mauromorales ! that fixed it!

mauromorales (approved):

PR #151, opened 2024-02-23
Reduce doc sizes to make kairos-agent smaller

mauromorales:
In total I manage to reduce the size from 15M to 8.1M (requires to run the cleanup script). If this is still too much for the binary we can stop including them, but it's still useful to do these fixes for maintenance and Google speed test

Additional:

- fix issue with descriptions including the entire document
- remove Google ID (since it was a sample ID but it was still loading the JS)

### DIFF
--- a/scripts/embedded-docs-cleanup.sh
+++ b/scripts/embedded-docs-cleanup.sh
@@ -3,4 +3,3 @@ location=${1:-public}
 
 find "$location" -type f -name "*.xml" -exec test -e {} \; -exec rm -f {} +
 find "$location" -type f -name "swagger-ui*" -exec test -e {} \; -exec rm -f {} +
-rm -rf "$location"/blog


hugo.toml
```diff
@@ -105,9 +106,6 @@ github_branch = "main"
 # Google Custom Search Engine ID. Remove or comment out to disable search.
 #gcs_engine_id = "011737558837375720776:fsdu1nryfng"
 
-# Enable Algolia DocSearch
-algolia_docsearch = false
```
jimmykarily, line 109: this is not what we are using for search right?

scripts/embedded-docs-cleanup.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
jimmykarily, line 1: who runs this script? It's in the kairos-agent side?

scripts/embedded-docs-cleanup.sh
```diff
@@ -0,0 +1,6 @@
...
+location=${1:-public}
+
+find "$location" -type f -name "*.xml" -exec test -e {} \; -exec rm -f {} +
+find "$location" -type f -name "swagger-ui*" -exec test -e {} \; -exec rm -f {} +
+rm -rf "$location"/blog
```
jimmykarily, line 6: does this mean we get 404s when a blog link is clicked?

jimmykarily (approved):
We don't save much space but if it works, let's do it.

hugo.toml
mauromorales: nope

scripts/embedded-docs-cleanup.sh
mauromorales: will need to be run in the package of the kairos-agent where we embed the docs

scripts/embedded-docs-cleanup.sh
mauromorales: ah this was just an experiment, good catch thanks!

PR #157, opened 2024-03-07
Expand docs about troublesshiting and insert an UKI section

Itxaka:


mudler:
I'd also add a reference to this section in the Installation https://kairos.io/docs/installation/trustedboot/ (also about the reset, so we can keep a starting point for users where to point when referring to trusted boot)

mudler (approved):

mauromorales (approved):

mauromorales:
we can now also cherry pick this into v3.0.0

mauromorales:
@Itxaka yup indeed

PR #163, opened 2024-03-18
trustedboot: example for installing files into the encrypted portion

mudler:
See: https://github.com/kairos-io/kairos/issues/2304#issuecomment-1988766974

### DIFF
--- a/content/en/docs/Installation/trustedboot.md
+++ b/content/en/docs/Installation/trustedboot.md
@@ -243,9 +243,74 @@ install:
   - PARTITION_TWO
 ```
 
-### Notes
+## Notes
 
-#### Mount partitions after install
+### Install additional files from the Installer ISO to the encrypted portions
+
+The ISO installer images can be used to install additional content in the encrypted portion of the disk. 
+
+The osbuilder image can overlay additional files into the iso. For example specify `--overlay-iso /additional/path` to have added the files in the folder inside the ISO. The content can be accessed during installation in `/run/initramfs/live`. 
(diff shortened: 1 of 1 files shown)


content/en/docs/Installation/trustedboot.md
```diff
@@ -243,9 +243,74 @@ install:
...
+
+users:
+- name: "kairos"
+  passwd: "kairos"
+
+stages:
+  after-install:
+      - commands:
+        - echo "Copying files to oem and persistent"
+        - /usr/lib/systemd/systemd-cryptsetup attach persistent $(findfs PARTLABEL=persistent) - tpm2-device=auto
```
Itxaka, line 274: I wonder if we should create a small subcommand to do this (kairos-agent lock/unlock LABEL) so its easier to do it if we expect this workflow to be typical.

Itxaka:
Wondering. For this to work under reset....you need the original media inserted right?

Sounds like we need a COS_STORAGE partition to store stuff in this case... coudl ship it in the media under whatever dir we want (extra-data or something) and copied and created to that partition. Then it would be available for reset/upgrade.

We could even have it encrypted and ONLY unlock it during reset/upgrade so its safe during runtime.

jimmykarily:
> Wondering. For this to work under reset....you need the original media inserted right?
> 
Nope, at install time, we write the data both to persistent and oem and at reset time, we copy them from oem back to persistent. We decided not to introduce another partition for simplicity. He used the `oem` one for that purpose.

jimmykarily (approved):

PR #178, opened 2024-04-10
fix: image shortcode

wcrum:
Current documentation points to https://github.com/kairos-io/kairos/releases/download/v3.0.4/kairos-opensuse-leap-15.5-standard-amd64-generic-v3.0.4-k3sv1.29.3-k3s1.iso here, https://kairos.io/docs/getting-started/#download. 

This is because when trying to replace `+` with `-`, the endpoint made does not exist.

Since `k3s_version` is equal to `v1.29.3+k3s1`, keeping `+` would output it as `v1.29.3&#43;k3s1` which is also incorrect, the way I fix this is by adding `safeHTML`, https://gohugo.io/functions/safe/html/.

### DIFF
--- a/layouts/shortcodes/image.html
+++ b/layouts/shortcodes/image.html
@@ -27,7 +27,7 @@ kairos-
 -
 {{- $.Page.Site.Params.softwareVersions.kairos_version -}}
 {{- if eq (.Get "variant") "standard" -}}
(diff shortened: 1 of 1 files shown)


mauromorales (approved):
Thanks @wcrum !

mauromorales:
I think we might have more places were we do this mistake :P thanks a lot for pointing it out!

PR #179, opened 2024-04-12
✨  Refactor homepage

mauromorales:
- remove unnecessary JS in favor of HTML
  - Our content was only rendered via JS which means SEO bots don't get to read it
  - The extra complexity is not needed for example, it makes it really complex to not have a link for fedora arm which is what html is good for
  - we cannot use the benefits of the asset pipeline of hugo
- load variables css
  - this means that some styles were not applied e.g. links with underlines when they were clearly not expected
- move images into assets
  - this allows to use the asset pipeline
  - found that some images which should not be images but just `<i>` with font awesome to avoid loading things more than once


Itxaka:
> @Itxaka you mean the download ones on the homepage? or which?

yep, the ones in the homepage (kairos.io) for arm64 they link to nowhere

jimmykarily (approved):

Itxaka (approved):

mauromorales:
@Itxaka oki will fix those, thanks for spotting it

PR #182, opened 2024-04-17
Document the use of the cmdline flags of enki

jimmykarily:
Documenting: https://github.com/kairos-io/enki/pull/88


content/en/docs/Installation/trustedboot.md
```diff
@@ -121,6 +121,44 @@ title My Awesome OS
 Remember when you build upgrade mediums, to also add your branding, otherwise the new active image will be named differently than your fallback and recovery ones.
 {{% /alert %}}
 
+### Additional efi entries
+
+When booting in UKI mode, the cmdline is part of the signed artifact. This means that in order to pass additional options through the cmdline,
+an additional efi entry is needed. `enki build-uki` supports the following arguments:
```
mauromorales, line 127: I don't think "an additional efi entry is needed" in all cases, like the extend one

content/en/docs/Installation/trustedboot.md
```diff
@@ -121,6 +121,44 @@ title My Awesome OS
...
+
+- `--extend-cmdline`: Use this option to add options to the cmdline used for active/passive/recovery. This option will not generate additional entries in the systemd-boot menu.
```
mauromorales, line 129: maybe use "attributes" or "stanzas" instead of the "to add options". 

Also since we are a the extended docs maybe we should also include "auto reset" in the active/passive/recovery. And maybe the best would be to describe that above and call this group of efi files something e.g. Kairos default entries

content/en/docs/Installation/trustedboot.md
```diff
@@ -121,6 +121,44 @@ title My Awesome OS
...
+
+Examples:
+
+Building with a command like this:
+
+```
+enki build-uki ...more options... --single-efi-cmdline "Lets debug: rd.debug rd.immucore.debug"
+```
+
+Will create a boot menu with the default entries, plus one more: "Lets debug" which will have the debug options set.
```
mauromorales, line 142: yeah here we use "default" entries and sounds good, so I think it would be safe to call it Kairos default entries

content/en/docs/Installation/trustedboot.md
jimmykarily: Good point. Re-phrasing.

content/en/docs/Installation/trustedboot.md
jimmykarily: I think the right term is "parameters" (https://www.kernel.org/doc/html/v4.14/admin-guide/kernel-parameters.html). Let me try to re-phrase.

content/en/docs/Installation/trustedboot.md
```diff
@@ -121,6 +121,44 @@ title My Awesome OS
...
+- `--extra-cmdline`: Use this option to generate additional entries in the systemd-boot menu. For each for active/passive/recovery, a new entry will be created with the specified cmdline (added to the default cmdline).
+- `--single-efi-cmdline`: Use this option to generate entries that are not related to active/passive/recovery. This option will generate a single entry with the provided cmdline (added to the default cmdline). This argument can be used multiple times.
+
+
+Examples:
+
+Building with a command like this:
+
+```
+enki build-uki ...more options... --single-efi-cmdline "Lets debug: rd.debug rd.immucore.debug"
```
mauromorales, line 139: TBH I'd really like to push back a bit with this way of setting the title of the single entry. I find it very confusing that for ALL other entries we use a single `--boot-branding` and now for this one, we have this special setup. So what happens when I do:

```
--boot-branding="AwesomeOS" --extend-cmdline="foo" --single-efi-cmdline="Debug: rd.debug"
```

Then I will have:

- AwesomeOS
- AwesomeOS (fallback)
- AwesomeOS recovery
- AwesomeOS reset (auto)
- Debug

content/en/docs/Installation/trustedboot.md
mauromorales: Also in the examples below, I think it would be nice to show the list of files generated and the boot menu, that will make it completely clear

mauromorales: parameters is good 👍

content/en/docs/Installation/trustedboot.md
mauromorales: @jimmykarily I think we should go back to product or at least the edge team, they requested this feature so they should be the one telling us if the branding is also expected on the single efi entry, but IMO branding should impact all entries including the single ones.

I don't think having to match flags is going to be tricky in the code, but maybe I'm missing something? Otherwise, then maybe it should actually be named `--single-efi-config=./path/to/my-efi.config` where `my-efi.config` includes: title, cmdline, more?

jimmykarily: @vipsharm any thoughts?

content/en/docs/Installation/trustedboot.md
```diff
@@ -121,6 +121,43 @@ title My Awesome OS
 Remember when you build upgrade mediums, to also add your branding, otherwise the new active image will be named differently than your fallback and recovery ones.
 {{% /alert %}}
 
+### Additional efi entries
```
mudler, line 124: to not confuse the user here, I'd move this down to Reference/Note.

I'd keep on the first part just the "minimum to follow", this looks like additional/advanced usage that a user might need after taking confidence with the process

content/en/docs/Installation/trustedboot.md
jimmykarily: Vipin said they'd like the branding to also apply to the "single" entries. This doesn't answer the question of how we set the unique part of the title though. So the user still needs to specify some text to be appended to the title. E.g. With `boot-branding` set to "My OS", the user needs a way to produce entries like:

```
My OS (debugging)
My OS (reset)
```

With a small modification in the current code, we can achieve this with:

```
enki build-uki ... --boot-branding "My OS" --single-efi-cmdline "debugging: rd.debug" --single-efi-cmdline "reset: someresetoption"
```

If we were to split the title and cmdline parts, we would need something like:

```
enki build-uki ... --boot-branding "My OS" --single-efi-cmdline "rd.debug" --single-efi-cmdline "someresetoption" --single-efi-title "debugging" --single-efi-title "reset"
```

and then we match the first cmdline with the first title etc. What happens when the user provides less/more titles than cmdlines? We can show a validation error, like we do when the title part is missing in the current implementation.

@mauromorales do you think this is better UX? I don't have a strong preference.

mauromorales: @jimmykarily if the branding applies to everything, I'm also ok with the original approach of `--single-efi-cmdline "debugging: rd.debug"` because I think it will be less likely that someone will want to put a more complex name there. So let's go with your first option. One last question, what will happen if I don't pass a name? Keep in mind that systemd-boot has its own way of differientiating boot entries that are named the same

jimmykarily: ok let me try things

jimmykarily: This: 

```
--boot-branding "KairosUKI" --single-efi-cmdline "My entry: someoption=somevalue" --single-efi-cmdline "Other entry: otheroption=othervalue" 
```

now produces:
image

I don't know if we want to do something about the ordering.

jimmykarily: If we were to allow empty titles for the single entries, this would happen:

image

(I hardcoded the title to "samething" to trigger this). Let's allow this then.

mauromorales: yeah I think the only problem with allowing that is that the string becomes so long (because of the long branding name used by product) that they cannot even read the last parts, but they might as well just name it something different :P

content/en/docs/Installation/trustedboot.md
jimmykarily: You mean I should move this whole section? Which part should I keep here?

jimmykarily: Done, moved

mauromorales (approved):

PR #198, opened 2024-05-17
landing_page: change wording and welcome text

mudler:


layouts/partials/content.html
```diff
@@ -236,4 +236,12 @@ <h3>Examples</h3>
             </a>
         </div>
     </div>
+    <div class="cncf-blade">
+        <div class="join-us-wrap">
+            {{ $cncf := resources.Get "/img/logo_cloudnative.png" }}
+            <h2>We are a Cloud Native Computing Foundation sandbox project.</h2>
+            <img src="{{ $cncf.RelPermalink }}" alt="opened box" loading="lazy"/>
+            <p>The Linux Foundation® (TLF) has registered trademarks and uses trademarks. For a list of TLF trademarks, see <a href="https://www.linuxfoundation.org/trademark-usage/" target="_blank">Trademark Usage</a>.</p>
```
mauromorales, line 244: IMO the color of the paragraph is not the best for readability, I think keeping it white would be better

mudler:
@mauromorales good points, let's tackle those in a follow up so meanwhile we get this going

mauromorales:
@mudler how about also putting a small CNCF logo on at the top of the page?

PR #200, opened 2024-05-22
Remove unused osbuilder flags

mauromorales:
fixes kairos-io/kairos#2212


Itxaka (approved):
Do we need to check the OCI vs docker part? Like the transport will tackle OCI as remote and docker as local IIRC? Or I'm remembering it wrong?

mauromorales:
@Itxaka good question, I'm not sure about that but regardless of, the local flag doesn't do anything

mauromorales:
@Itxaka ah I'm silly this is the docs so you mean to inform the user, my bad sorry. I will test and follow up

PR #207, opened 2024-06-04
Add instructions on how to install the system-upgrade-controller

jimmykarily:
before showing the example that needs it.

Part of: https://github.com/kairos-io/kairos/issues/2586

### DIFF
--- a/content/en/docs/Upgrade/trustedboot.md
+++ b/content/en/docs/Upgrade/trustedboot.md
@@ -71,6 +71,7 @@ When invoking `kairos-agent` in the plan however, you need to specify the `--sou
 
 In the following example `<CONTAINER_IMAGE>` is the source image used to generate the upgrade image, `<CONTAINER_IMAGE_TAG>` is the tag of that image and `<UPGRADE_IMAGE>` is the generated upgrade image (tag included) as per the documentation above.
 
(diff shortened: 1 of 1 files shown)


jimmykarily:
This is one way to fix https://github.com/kairos-io/kairos/issues/2586. We don't merge the 2 suc-upgrade scripts because the path where the efi files reside depends on [how the user creates the "plan" image](https://github.com/kairos-io/kairos-docs/pull/207/files#diff-06d4b5f4dfdd276ca8e2c5e73668fa2040f5a94988b7dc1e9ed10a68ccab5f33R95).

An alternative way would be to make enki build an image suitable for a Plan Pod but if that image is passed to `kairos-agent upgrade --source oci:Image_here` , it will need changes in the kairos-agent to only extract the efi files and not the whole OS from the image.

In my opinion, going with docs for this is simpler. We can go with the second option in the future if we decide we didn't like it after all.

@kairos-io/maintainers ?

mauromorales (approved):

jimmykarily:
My comment above was meant to go to this PR: https://github.com/kairos-io/kairos-docs/pull/209 :facepalm: 

In any case, let me merge the 2 PRs together in one because they are relevant.

jimmykarily:
> My comment above was meant to go to this PR: #209 🤦
> 
> In any case, let me merge the 2 PRs together in one because they are relevant.

nope, it's just morning and I'm not awake yet. The comment and the PRs are just fine, these are 2 different issues. Sorry for the noise.

PR #209, opened 2024-06-11
Add general information about revocation

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/2429


content/en/docs/Advanced/revoking-secureboot-access.md
```diff
@@ -0,0 +1,300 @@
...
+```
+
+
+#### Upgrade to the new image
+
+```
+kairos-agent upgrade --source oci:myimage
+```
+
+#### Unbind PCR 7
```
Itxaka, line 167: I think the ordering is wrong here no?  You mentioned above

- We "unbind" decryption from PCR 7, binding only to PCR 11 indirectly.
- We enroll the blacklisted cert in dbx.
- We reboot to the upgraded system.
- We bind decryption again both to PCR 11 and 7 (as before).

But here we are  doing it on a different order in the docs. Would it make sense to have it in the correct order?

content/en/docs/Advanced/revoking-secureboot-access.md
```diff
@@ -0,0 +1,300 @@
...
+
+Bind partition encryption only to PCR 11 policy. From within Kairos again:
+
+```bash
+echo "Generating temporary passphrase"
+dd if=/dev/random bs=32 count=1 of=/tmp/random_keyfile
+
+echo "Adding password slot"
+cryptsetup luksAddKey --token-type systemd-tpm2 /dev/vda2 /tmp/random_keyfile
```
Itxaka, line 176: maybe we should explain about this?

content/en/docs/Advanced/revoking-secureboot-access.md
```diff
@@ -0,0 +1,300 @@
...
+cryptsetup luksAddKey --token-type systemd-tpm2 /dev/vda3 /tmp/random_keyfile
+
+echo "Removing the tpm2 slot"
+systemd-cryptenroll /dev/vda2 --wipe-slot=tpm2
+systemd-cryptenroll /dev/vda3 --wipe-slot=tpm2
+
+echo "Adding tpm2 slot again (pcr 11 policy AND pcr 7)"
+echo "There is probably a bug in systemd preventing us from using the --tpm2-device-key"
```
Itxaka, line 219: Weird. It should have the public key in there always?

Itxaka (approved):
nocie, just a couple of comments

content/en/docs/Advanced/revoking-secureboot-access.md
```diff
@@ -0,0 +1,300 @@
...
+title: "Revoking secure boot access"
+linkTitle: "Revoking secure boot access"
+weight: 10
+date: 2022-06-12
+---
+
+This document describes how an administrator can prevent certain OS images from
+booting on their hardware in the context of ["Trusted Boot"]({{< relref "../architecture/trustedboot" >}}).
+
+Two different scenarios will be covered, with the process being only slightly different for each case.
```
mauromorales, line 11: it talks about two scenarios but I only see one bellow

content/en/docs/Advanced/revoking-secureboot-access.md
```diff
@@ -0,0 +1,300 @@
...
+
+## Scenario 1 - Signing certificate no longer trusted
```
mauromorales, line 13: what does this actually mean, that you sign a certificate which is no longer trusted? or that a signed certificate is no longer trusted?

content/en/docs/Advanced/revoking-secureboot-access.md
```diff
@@ -0,0 +1,300 @@
...
+
+The process of creating signed images that can be trusted to boot, requires the
+signing keys to be safe and only accessible to the vendor that produces the OS images.
+
+In the unfortunate case in which a signing certificate has been compromised, it's
+important that the keys is blacklisted it on every machine.
+Failing to do so, will allow be possible for the malicious actor with access to the key,
```
mauromorales, line 20:
```suggestion
Failing to do so, will make it possible for the malicious actor with access to the key,
```

content/en/docs/Advanced/revoking-secureboot-access.md
```diff
@@ -0,0 +1,300 @@
...
+to generate OS images that the system will accept to boot.
+This will allow them boot and decrypt the filesystem, thus gaining full access
+to the data on that machine and full control over it.
+
+The UEFI firmware has a special "database" in which blacklisted certificates can
+be "enrolled" (image hashes can also be enrolled, but more on that on the next scenario).
+
+If only one machine is running the signed OS images and physical access to that machine is
```
mauromorales, line 28:
```suggestion
If only one machine is running the signed OS images, and physical access to that machine is
```

content/en/docs/Advanced/revoking-secureboot-access.md
```diff
@@ -0,0 +1,300 @@
...
+
+There are various utilities in Linux that allow enrolling certificates in the UEFI
+databases but they require those certs to be in a specific format usually in the
+ESL (EFI Signature List) or the signed version of the same format, the "signed variables"
+format. Using the signed version has the benefit that no private keys need to be present
+in order to do the enrollement, while using the unsigned format needs the private
+keys to be present.
+
+If you followed [the instructions to create signing keys]({{< relref "../Installation/trustedboot.md#key-generation" >}}), you should have a directory with a `db.pem` certificate. This is the certificate
+we will blacklist. To do so we need to convert it to the "esl" format and then
```
mauromorales, line 53:
```suggestion
we will blacklist. To do so ,we need to convert it to the "esl" format and then
```

content/en/docs/Advanced/revoking-secureboot-access.md
```diff
@@ -0,0 +1,300 @@
...
+sign it to create the final `.auth` file which will be used for enrollement.
```
mauromorales, line 54:
```suggestion
sign it to create the final `.auth` file, which will be used for enrollement.
```

content/en/docs/Advanced/revoking-secureboot-access.md
```diff
@@ -0,0 +1,300 @@
...
+
+As described in the ["Trusted Boot"]({{< relref "../architecture/trustedboot" >}}) documentation page,
+the decryption of the disk partitions is bound to some PCR registers on the TPM chip.
+Specifically registers 11 and 7.
+There are 2 ways to bind to a PCR register, the direct and the indirect. You can
+read more in the [`systemd-cryptenroll` docs](https://www.freedesktop.org/software/systemd/man/latest/systemd-cryptenroll.html#TPM2%20PCRs%20and%20policies) but in a nutshell, the indirect binding
+allows the actual value of the PCR to change while the direct one does not.
+For this reason, in Kairos, the decryption is bound indirectly to PCR 11 (which allows
+us to upgrade to newer kernels) and directly to PCR 7, which prevents booting if
+the UEFI databases have been altered, e.g. by enrolling a new key.
```
mauromorales, line 75: No need to do this change but it sounds quite important so I'd put on bold

```suggestion
**For this reason, in Kairos, the decryption is bound indirectly to PCR 11 (which allows
us to upgrade to newer kernels) and directly to PCR 7, which prevents booting if
the UEFI databases have been altered, e.g. by enrolling a new key.**
```

content/en/docs/Advanced/revoking-secureboot-access.md
```diff
@@ -0,0 +1,300 @@
...
+
+From withing Kairos (still booted in the old image), enroll the new db key:
+
+```bash
+sudo chattr -i /sys/firmware/efi/efivars/{PK,KEK,db}*
+efi-updatevar -f db.auth db
+```
+
+{{% alert title="Warning" %}}
+You will need to `efitools` installed on the Kairos image for this commands to work!
```
mauromorales, line 147:
```suggestion
You will need the`efitools` installed on the Kairos image for this commands to work!
```

content/en/docs/Advanced/revoking-secureboot-access.md
jimmykarily: The latter. Maybe ` Signing certificate is no longer trusted` ?

content/en/docs/Advanced/revoking-secureboot-access.md
jimmykarily: Good point. Actually the order between enrolling to dbx and upgrading doesn't matter. We just can't reboot before we unbind from PCR 7. But I'll re-order to avoid confusing people.

content/en/docs/Advanced/revoking-secureboot-access.md
```diff
@@ -0,0 +1,300 @@
...
+
+#### Unbind PCR 7
+
+Bind partition encryption only to PCR 11 policy. From within Kairos again:
+
+```bash
+echo "Generating temporary passphrase"
+dd if=/dev/random bs=32 count=1 of=/tmp/random_keyfile
+
+echo "Adding password slot"
```
jimmykarily, line 175:
```suggestion
echo "Adding password slot"
# We temporarily switch to a passphrase decryption here, so that we can remove the "tpm" decryption
# option below. After we add the "tpm" option back (bound to different PCR registers),
# we remove the passphrase again.
```

content/en/docs/Advanced/revoking-secureboot-access.md
jimmykarily: How about the above ?

content/en/docs/Advanced/revoking-secureboot-access.md
jimmykarily: where? Are you referring to the commands below? I pass the public key, yes. Do you mean I shouldn't?

content/en/docs/Advanced/revoking-secureboot-access.md
mauromorales: yup sounds good to me

content/en/docs/Advanced/revoking-secureboot-access.md
jimmykarily: I pushed a commit. I tried to refer to scenario 1 and only present the differences to avoid duplicating the whole thing.

content/en/docs/Advanced/revoking-secureboot-access.md
```diff
@@ -0,0 +1,337 @@
...
+
+In this scenario, only one specific image is considered "bad". For example, an image shipped with a well known security vulnerability.
+In order to prevent this image from booting, its "hash" can be blacklisted.
+
+The process is exactly the same as above, but instead of enrolling a certificate in the `dbx` database, we will enroll an image hash.
+
+Follow the same instructions as in scenario 1, but skipping the steps where a new `db` key is created.
+
+The easiest way to find the offending `.efi` file and calculate its signature is to boot the image which we
+want to blacklist in a sandboxed environment.
```
jimmykarily, line 314: I'm not sure we should suggest this. Maybe I should show instructions for the extraction from an ISO. But it gets messy.

content/en/docs/Advanced/revoking-secureboot-access.md
```diff
@@ -0,0 +1,337 @@
...
+weight: 10
+date: 2022-06-12
+---
+
+This document describes how an administrator can prevent certain OS images from
+booting on their hardware in the context of ["Trusted Boot"]({{< relref "../architecture/trustedboot" >}}).
+
+Two different scenarios will be covered, with the process being only slightly different for each case.
+
+## Scenario 1 - Signing certificate no longer trusted
```
jimmykarily, line 13:
```suggestion
## Scenario 1 - Signing certificate is no longer trusted
```

content/en/docs/Advanced/revoking-secureboot-access.md
Itxaka: in the machine I mean, you mention a bug preventing from using the device key, but its not clear to me what the issue is or why using the tpm-device instead makes it work?

skipping hitting the tpm device should be the way to go always, so we prevent blocking the tpm AND we can choose the PCR numbers that we want. Using --device-auto should be the last resort only done on systemd <255

jimmykarily: Well, it was you that suggested that this is a bug in systemd, remember? :D And actually you suggested the command that works.

Itxaka: yes, but thats for enrolling when we DONT want to bind to PCR7, when we want to do so, we can use the 

Basically:

 - using` --tpm2-device` makes it so the PCR7 is always binded no matter what flags you use
 - using `--tpm2-device-key` makes it so we can bind to anything, even leave PCR bind empty
 - using `--tpm2-device-key` also does not hit the TPM at all, which is nice as we can end up blocking it
 - `--tpm2-device-key` can only be used on systemd 255 (buntu 24.04 and Fedora 40)

So we should favor using always the `--tpm2-device-key` flag with its other `--tpm2-pcrs` and `--tpm2-public-key-pcrs` flags to have full control over whats binding or not.

And because we need to use that `--tpm2-device-key` in order to unbind from PCR7 we should only mention that flag and ignore the `--tpm2-device` one

content/en/docs/Advanced/revoking-secureboot-access.md
Itxaka: sounds good!

content/en/docs/Advanced/revoking-secureboot-access.md
Itxaka: yes indeed. Maybe on EFI build we should spit a file which contains info about the EFI image AND its ginature+hash and so on? So that can get archived somewhere in order to refer to it? So then you can just use that hash to blacklist it or know which signature it was used?

Itxaka: also maybe suggest that you can boot with the newer install media, and calculate it via that, so you are on a safe environment? boot from livecd, mount efi partition, calculate signatures or whatever

content/en/docs/Advanced/revoking-secureboot-access.md
```diff
@@ -0,0 +1,340 @@
...
+
+From withing Kairos (still booted in the old image), enroll the new db key:
+
+```bash
+sudo chattr -i /sys/firmware/efi/efivars/{PK,KEK,db}*
+efi-updatevar -f db.auth db
+```
+
+{{% alert title="Warning" %}}
+You will need the`efitools` installed on the Kairos image for this commands to work!
```
jimmykarily, line 147:
```suggestion
You will need the `efitools` installed on the Kairos image for this commands to work!
```

content/en/docs/Advanced/revoking-secureboot-access.md
jimmykarily: This is the error I'm getting:

```
root@localhost:/home/kairos# systemd-cryptenroll --unlock-key-file=/tmp/random_keyfile --tpm2-public-key=/run/systemd/tpm2-pcr-public-key.pem --tpm2-public-key-pcrs=11 --tpm2-pcrs=7 --tpm2-signature=/run/systemd/tpm2-pcr-signature.json --tpm2-device-key=/run/systemd/tpm2-srk-public-key.tpm2b_public /dev/vda2
/tmp/random_keyfile has 0644 mode that is too permissive, please adjust the ownership and access mode.
Must provide all PCR values when using TPM2 device key.
```

content/en/docs/Advanced/revoking-secureboot-access.md
jimmykarily: Producing the information at build time (your first suggestion) doesn't sound enough. Maybe they no longer have access to those files anymore (the build could have happened months ago).

I prefer your second suggestion because it doesn't require access to any old files. But I need to try that out and come up with instructions. Let me try.

jimmykarily: I added some instructions around this (not copy-paste things)

content/en/docs/Advanced/revoking-secureboot-access.md
jimmykarily: re-ordered

Itxaka (approved):

content/en/docs/Advanced/revoking-secureboot-access.md
jimmykarily: ok, we decided to keep it as is for now.

PR #227, opened 2024-07-22
Fix blog entry header

Itxaka:


Itxaka:
merging this directly!

PR #229, opened 2024-07-25
Upgrade docs improvements

mauromorales:
- add the suc version to a variable so we can automatically update it
- remove the recovery section of upgrades and put it together with the manual section
- add descriptions to all the sections
- reorder


content/en/docs/Upgrade/manual.md
```diff
@@ -37,6 +42,22 @@ For example, if you wanted to upgrade to the latest available stable release you
 sudo kairos-agent upgrade --source oci:{{<oci variant="standard">}}
 ```
 
-{{% alert title="Note" %}}
-Looking to upgrade from a private registry OCI image? Check the [Private registry auth]({{< relref "../Advanced/private_registry_auth" >}}) page.
+Once you have tested the new system and are happy with it, you can upgrade the recovery system.
+
+## Upgrading the recovery system
+
+The recovery system is there for a reason, to help you recover the active system in case of failure. This is why we don't allow upgrading the active system and recovery one at the same time. You should try to keep your recovery system as close as possible to the active one.
```
jimmykarily, line 49: I'm not sure about the last sentence. Why is it important that the recovery is close to the active? It may create the urge to the user to upgrade recovery as quickly as possible but how long should the active system be running before they consider it's ok to bump recovery to it? Maybe we should loosen it up a bit with something like:

"It's advised to also upgrade the recovery system often, to keep it close to the active one. This will make sure you have a familiar system to work with, when you boot to the recovery system instead of an old image you haven't used for quite a long time"

jimmykarily (approved):

content/en/docs/Upgrade/manual.md
mauromorales: yup sounds better, will update

PR #237, opened 2024-08-07
Add the ability to select & save prefered distro

mauromorales:


assets/js/kairos.js
```diff
@@ -0,0 +1,161 @@
...
+        }
+    }
+    return null; // If the text is not found
+}
+
+function calculateBaseImage(distroInfo) {
+    let baseImage = distroInfo[1] + ':' + distroInfo[2];
+
+    if (distroInfo[1] == 'opensuse') {
+        baseImage = distroInfo[1] + '/' + distroInfo[2].replace('-', ':');
```
Itxaka, line 18: Can you add a comment on why do we have to do this for opensuse?

content/en/docs/Advanced/build.md
```diff
@@ -122,12 +122,12 @@ $ curl http://<node-ip>:$PORT/hello-kairos.iso -o output.iso
 
 It is possible to use the CRD to prepare artifacts required for netbooting, by enabling `netboot: true` for instance:
 
-```yaml
+```yaml {class="meta-distro"}
```
Itxaka, line 125: is it possible to override this template via the base templates somehow to always append the this class? Otherwise this will only works if we remember to add the class manually no? 

IIRC this ` ```LANGUAGE` stuff is provided via a template or function so we may be able to extend it directly in there?

Itxaka (approved):
looking good to me!

content/en/docs/Advanced/build.md
mauromorales: yeah good point, let me see what I can find

mauromorales: refactored

PR #242, opened 2024-08-25
Update single-node.md

jimmyjones2:
Removed some spurious copied config and text - traefik and servicelb should be enabled on a single node cluster and remove reference to write_files as isn't included in config

### DIFF
--- a/content/en/docs/Examples/single-node.md
+++ b/content/en/docs/Examples/single-node.md
@@ -28,12 +28,29 @@ users:
 k3s:
   enabled: true
   args:
-  - --disable=traefik,servicelb
+  - --node-label "nodetype=small"
+
+write_files:
+- path: /var/lib/rancher/k3s/server/manifests/myapp.yaml
+  permissions: "0644"
+  content: |
+    apiVersion: v1
+    kind: Namespace
+    metadata:
+      name: myapp
+    ---
+    apiVersion: v1
+    kind: ConfigMap
+    metadata:
+      name: foobar
+      namespace: myapp
+    data:
+      foo: bar
 ```
 
 Notably:
 
-- We use the `k3s` block to disable `traefik` and `servicelb` (the default `k3s` load balancer).
+- We use the `k3s` block to set the node label
 - In a single-node setup, you may wish to use a non-generated node name. This can be achieved with these options:
   ```
   k3s:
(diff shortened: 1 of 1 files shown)


jimmykarily:
Hi @jimmyjones2 , thanks for this PR. I think the k3s flags were there only as an example on how to pass config to k3s, not as a recommendation on which flags to pass. The write_files must have been removed at some point from the config but I'd rather add an example of a yaml file to write than removing that altogether.

Maybe we can explain that the flags are there just for demonstration purposes of what can be done, what do you think?

jimmyjones2:
How about:
```
write_files:
- path: /var/lib/rancher/k3s/server/manifests/myapp.yaml
  permissions: "0644"
  content: |
    apiVersion: v1
    kind: Namespace
    metadata:
      name: myapp
    ---
    apiVersion: v1                                                          
    kind: ConfigMap                                                         
    metadata:                                                               
      name: foobar                                                          
      namespace: myapp                                                    
    data:                                                                   
      foo: bar
```

Can you suggest an example argument to k3s to include?

There is already an example in this page for setting the node name, but I think it would make more sense to remove that and update `hostname: myserver` which k3s will then use as the node name, rather than them being different?

jimmykarily:
> How about:
> 
> ```
> write_files:
> - path: /var/lib/rancher/k3s/server/manifests/myapp.yaml
>   permissions: "0644"
>   content: |
>     apiVersion: v1
>     kind: Namespace
>     metadata:
>       name: myapp
>     ---
>     apiVersion: v1                                                          
>     kind: ConfigMap                                                         
>     metadata:                                                               
>       name: foobar                                                          
>       namespace: myapp                                                    
>     data:                                                                   
>       foo: bar
> ```
> 
> Can you suggest an example argument to k3s to include?
> 
> There is already an example in this page for setting the node name, but I think it would make more sense to remove that and update `hostname: myserver` which k3s will then use as the node name, rather than them being different?

The example looks good. Regarding the k3s args, I think any of those mentions [in the docs](https://docs.k3s.io/installation/configuration) would do. Maybe `--node-label "foo=bar"`? It's meant to be an example anyway.

jimmyjones2:
@jimmykarily Have made the updates as requested, can you take another look please?

jimmykarily (approved):

jimmykarily:
All good, thanks @jimmyjones2 !

PR #243, opened 2024-08-25
Clarify kairos user can always sudo

jimmyjones2:
This is mentioned in getting started, but thought it would be useful to clarify here


mauromorales (approved):
Thanks for the PR

PR #244, opened 2024-08-28
Fix upgrade links

flesser:
those two links on https://kairos.io/docs/architecture/immutable/ did not work and threw 404 instead


mauromorales (approved):
thanks!

PR #249, opened 2024-09-03
Add Getting Started Configuration page

mauromorales:


content/en/docs/Getting started/configuration.md
```diff
@@ -0,0 +1,227 @@
...
+
+The `k3s` directive is used to enable or disable the Kubernetes distribution that comes with Kairos. If you set it to `false`, the K3s services will not start on boot.
+
+The `name` directive can be added at every level of the configuration file. It is useful to distinguish between different steps.
+
+The `stages` directive is used to define the different stages of the configuration. The `network` stage is used to configure the system when network starts. In this case, we create a user called `kairos` with the password `kairos` and an SSH key.
+
+## Configuring the hostname
+
+Configuring the hostname is a simple task. All we have to do is add the `hostname` directive at the root level of the configuration file with the value we want to give to this node. The order of the directives isot important. Let's add it after the `k3s` directive:
```
jimmykarily, line 100:
```suggestion
Configuring the hostname is a simple task. All we have to do is add the `hostname` directive at the root level of the configuration file with the value we want to give to this node. The order of the directives is important. Let's add it after the `k3s` directive:
```

content/en/docs/Getting started/configuration.md
```diff
@@ -0,0 +1,227 @@
...
+Let's take this a bit further. In the future we want to add a second master node to our cluster. Let's then use a value that is unique to this node and add it as a suffix.
+
+```yaml
+#cloud-config
+
+install:
+    device: /dev/vda
+k3s:
+    enabled: true
+hostname: "master-{{ trunc 4 .MachineID }}"
```
jimmykarily, line 138: Maybe we should point the users to some docs on what templating functions are available? To at least explain why this works.

jimmykarily (approved):
Just some minor comments.
Well needed page indeed, good job!

content/en/docs/Getting started/configuration.md
mauromorales: @jimmykarily very good point, I was looking for it on the docs but couldn't find a place where we define this, do you know where they come from?

mauromorales: found it http://masterminds.github.io/sprig/

jimmykarily: I wanna see you documenting this part :D https://github.com/mudler/yip/blob/3eeeb15bc6b921a977c1f44cf30e77c81dcd4dc6/pkg/plugins/hostname.go#L42

jimmykarily: I think user also gets access to these: https://github.com/zcalusic/sysinfo
because of this: https://github.com/mudler/yip/blob/3eeeb15bc6b921a977c1f44cf30e77c81dcd4dc6/pkg/plugins/common.go#L47

but I'm not 100% sure

PR #250, opened 2024-09-03
Add only flavors for customizing page

mauromorales:
The content of the docker file is specific to opensuse


mauromorales:
not what I wanted to do actually :P

PR #254, opened 2024-09-05
Add Getting Started build and upgrade guide

mauromorales:


jimmykarily:
[Minor] The order of link in the menu doesn't follow the linking in the pages themselves:

image

The "What is Kairos" page is only linked to by the top level "Getting started" page but in the menu it is sitting right between the 2 other links. Maybe make it the first "step" in the guide and adjust the other pages accordingly?

mauromorales:
> The "What is Kairos" page is only linked to by the top level "Getting started" page but in the menu it is sitting right between the 2 other links. Maybe make it the first "step" in the guide and adjust the other pages accordingly?

good point let me finish the "getting started" section, and I will reorder everything properly

PR #255, opened 2024-09-07
Update building-and-upgrading.md

tuxtof:
update Dockerfile and fix apt install command who was waiting input to validate install


mauromorales (approved):
Thanks for the fix!

PR #261, opened 2024-09-16
Explain what adding a user to the "admin" group does

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/2821


jimmykarily:
This is what it looks like: 
image

Itxaka:
I wonder why we dont use the wheel group as that seems to be more in line with most of the distros that use sudo... @mudler do you know?

Itxaka (approved):

mudler:
> I wonder why we dont use the wheel group as that seems to be more in line with most of the distros that use sudo... @mudler do you know?

IIRC the rationale behind it was to avoid mixups in configuration: for instance, a flavor (e.g. Ubuntu) could change defaults of wheel and we could get hit and have to jump into monkey patching for that specific flavor.

mudler (approved):

mauromorales (approved):

PR #262, opened 2024-09-16
Add building sysext with enki info

Itxaka:
Fixes: https://github.com/kairos-io/kairos/issues/2861

### DIFF
--- a/content/en/docs/Advanced/sys-extensions.md
+++ b/content/en/docs/Advanced/sys-extensions.md
@@ -75,7 +75,9 @@ This feature is in preview state and only available in Enki from version 0.1.4
 
 You can also build a system extension from a docker image directly by using [enki](https://github.com/kairos-io/enki) and using a dockerfile to isolate the artifacts you want converted into a system extension.
 
-Notice that when converting a docker image into a system extension, the last layer is the only one converted (The last command in a given Dockerfile) so have that in mind. This is useful for packages that ONLY install things in /usr or manual installation under /usr
+Notice that when converting a docker image into a system extension, the last layer is the only one converted (The last command in a given Dockerfile) so have that in mind. This is useful for packages that ONLY install things in /usr or manual installation under /usr.
+
+The `/usr/lib/extension-release.d/extension-release.NAME` file necessary for identifying the system extension is automatically created by the command so in this case you should not worry about that file.
 
 
 For example for a given Dockerfiles as such:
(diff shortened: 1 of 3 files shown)


mauromorales (approved):
lgtm

content/en/docs/Advanced/sys-extensions.md
```diff
@@ -62,8 +66,80 @@ This will generate a signed+verity sysextension that can then be used by sysext
...
+ - `-v /var/run/docker.sock:/var/run/docker.sock`: We pass the docker sock into the container so it can access our locally built container images. So we avoid pushing them and pulling them from a remote registry.
+ - `--rm`: Once the container exit, remove it so we dont leave stuff lying around.
+
+The explanation of the Enki command flags is as follows:
+ - `sysext`: Subcommand to call, in this case we want to build a sysext
+ - `NAME`: Output and internal name of the sysext.
+ - `CONTAINER_IMAGE`: Image from which we will extract the last layer and covert it to a system extension.
+ - `--private-key`: Private key to sign the system extension.
+ - `--certificate`: Certificate to sign the system extension.
+ - `--output`: Dir where we will output the system extension. Make sure that this matches what is passed to the docker command to be able to keep the generated system extension once the container exists and its removed.
```
jimmykarily, line 126:
```suggestion
 - `--output`: Dir where we will output the system extension. Make sure that this matches the directory that passed to the docker command to be able to keep the generated system extension once the container exists and its removed.
```

content/en/docs/Advanced/sys-extensions.md
```diff
@@ -62,8 +66,80 @@ This will generate a signed+verity sysextension that can then be used by sysext
...
+
+```dockerfile
+FROM anchore/grype:latest AS grype
+
+
+FROM scratch
+COPY --from=grype /grype /usr/local/bin/grype
+```
 
+Only the files added in the last step will be converted to a sysext, so the contents of the sysext would be the `/usr/local/bin/grype` binary only.
```
jimmykarily, line 91: Not really. The file with the content `id=_ANY` is auto-generated. We'd better mention that here because while reviewing the PR I didn't know that and it makes is a real pain to create that file too with heredoc in the dockerfile and still keep everything in one command (last layer).

jimmykarily (approved):

content/en/docs/Advanced/sys-extensions.md
Itxaka: Good point. I did not add it as it seems superfluous to me, but the point that you created the file yourself without knowing that the command would create it automatically says that this info is probably needed here. Will add it somewhere around.

PR #263, opened 2024-09-17
Replace install.image with install.source in docs

jimmykarily:
because the old option is deprecated

Fixes https://github.com/kairos-io/kairos/issues/2810


Itxaka (approved):

mauromorales (approved):

content/en/docs/Examples/core.md
paddy-hack, line 1: On line 14 there still is a `install.image` and on line 42 an `image:` :sob: 

You may want to run (something like)
``` sh
git grep -E '(install\.image|image:)'
```
to check for more.

PR #264, opened 2024-09-17
Create docs page

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/2798


jimmykarily:
This PR addresses most of the items in the ticket description but I'm not sure what was meant by "Check image integrity". I couldn't find any other mechanisms we have in place. If I missed something let me know and I will add it.

jimmykarily:
> looks good. I wonder if we should add the same parameters to the trusted boot install defaults to have automatic reboots on kernel and systemd crashes...

We should. But this is the "documentation" ticket. I'm only documenting what we already have :D. I will create a ticket for this.

jimmykarily:
> > looks good. I wonder if we should add the same parameters to the trusted boot install defaults to have automatic reboots on kernel and systemd crashes...
> 
> We should. But this is the "documentation" ticket. I'm only documenting what we already have :D. I will create a ticket for this.

https://github.com/kairos-io/kairos/issues/2869

PR #265, opened 2024-09-18
Split stage modules in a new page are refer to it in 2 places

jimmykarily:
also re-order the modules in that page so that they match the order of execution.

Finally add documentation for systemd-firstboot.

Fixes: https://github.com/kairos-io/kairos/issues/2792


content/en/docs/Reference/stage_modules.md
```diff
@@ -0,0 +1,449 @@
...
+     - name: "Setup users"
+       environment:
+         FOO: "bar"
+```
+
+### `systemd_firstboot`
+
+Runs [`systemd-firstboot`](https://www.freedesktop.org/software/systemd/man/latest/systemd-firstboot.html) with the arguments specified.
+
+```yaml
```
jimmykarily, line 335: I tried this example and it doesn't seem to have any effect. If I run `systemd-firstboot --hostname=myhostname` manually it creates `/etc/hostname` correctly. I can see in the agent logs that the stage was run but for some reason it had no effect.

Itxaka (approved):

PR #266, opened 2024-09-18
Expand the stages, add a specific picture for install process

Itxaka:
Fixes https://github.com/kairos-io/kairos/issues/2759


Itxaka:
I wonder if the stages stuff needs its own page linked from here

content/en/docs/Architecture/cloud-init.md
```diff
@@ -41,32 +41,51 @@ stages:
...
+| _kairos-uki-reset.pre_     | This stage is executed before reset. Only run under Trusted Boot                                                                                                                                                                                                                    |                         
+| _kairos-uki-reset.after_   | This stage is executed after reset. Only run under Trusted Boot                                                                                                                                                                                                                     |                       
+| _kairos-uki-upgrade.pre_   | This stage is executed before reset. Only run under Trusted Boot                                                                                                                                                                                                                    |
```
jimmykarily, line 58:
```suggestion
| _kairos-uki-upgrade.pre_   | This stage is executed before upgrade. Only run under Trusted Boot                                                                                                                                                                                                                    |                       
```

content/en/docs/Architecture/cloud-init.md
```diff
@@ -41,32 +41,51 @@ stages:
...
+| _after-install-chroot_     | This stage happens after installing active and grub                                                                                                                                                                                                                                 |
+| _after-install_            | This stage runs after active,passive and recovery images are installed and after disks have been encrypted                                                                                                                                                                          |
+| _before-reset_             | This stage happens after partitions have been formatted and mounted but before the image has been resetted                                                                                                                                                                          |
```
jimmykarily, line 63:
```suggestion
| _before-reset_             | This stage happens after partitions have been formatted and mounted but before the image has been reset                                                                                                                                                                          |
```

content/en/docs/Architecture/cloud-init.md
```diff
@@ -41,32 +41,51 @@ stages:
...
+| _after-reset_              | This stage happens after partitions have been formatted and mounted and active and passive images resetted                                                                                                                                                                          |
```
jimmykarily, line 64:
```suggestion
| _after-reset_              | This stage happens after partitions have been formatted and mounted and active and passive images reset                                                                                                                                                                          |
```

content/en/docs/Architecture/cloud-init.md
```diff
@@ -41,32 +41,51 @@ stages:
...
+| _before-upgrade_           | This stage happens after mounting partitions with RW but before any image has been upgraded                                                                                                                                                                                         |
+| _after-upgrade_            | This stage happens after upgrade has been done                                                                                                                                                                                                                                      |
+| _after-upgrade-chroot_     | This stage happens after the image has been upgraded                                                                                                                                                                                                                                |
```
jimmykarily, line 68: what's the difference with the previous one? Maybe explain here?

content/en/docs/Architecture/cloud-init.md
```diff
@@ -41,32 +41,51 @@ stages:
...
-| _provider-kairos.bootstrap.after.<role>_               | The provider fires this stage after it finished bootstrapping K3S.                                                 |
+| **Stage**                                 | **Description**                                                    |
+|-------------------------------------------|--------------------------------------------------------------------|
+| _provider-kairos.bootstrap.before.<role>_ | The provider fires this stage before starting to bootstrap K3S.    |
+| _provider-kairos.bootstrap.after.<role>_  | The provider fires this stage after it finished bootstrapping K3S. |
+
+
+### Stages during kairos-agent operations in detail
+
+![Install Stages](https://github.com/user-attachments/assets/b050f247-990a-4395-9b45-334e04f84d45)
```
jimmykarily, line 88: image
does this mean that "Encrypt partitions" happens after "after-install" ? Maybe `after-install` should appear before `Encrypt partitions` box somehow? Something to make it more clear, I'm not a design expert for sure :D

jimmykarily (changes requested):

content/en/docs/Architecture/cloud-init.md
Itxaka: its explained at the end that all chroot stages happens in a chroot inside the affected image :D

content/en/docs/Architecture/cloud-init.md
Itxaka: yes this is difficult to explain. I tried to convey that after encrypt aprtitions, the `after-install` stage runs but it doesnt seems like its clear enough.

@mauromorales you got the best design vision I think, any ideas how to make this clearer?

content/en/docs/Architecture/cloud-init.md
mauromorales: @jimmykarily if you do the diff "without white space" it's a bit clearer to see what was changed, and what was added

content/en/docs/Architecture/cloud-init.md
```diff
@@ -41,32 +41,51 @@ stages:
...
+| _network_                  | This stage is executed when network is available                                                                                                                                                                                                                                    |
+| _reconcile_                | This stage is executed 5m after boot and periodically each 60m.                                                                                                                                                                                                                     |
+| _kairos-install.pre_       | This stage is executed before installation of the OS starts                                                                                                                                                                                                                         |
```
mauromorales, line 52: Strange that we have this, and also `_before-install_` I think I'd rather just have his one, and call the other one `_partitioning.pre_`. Is that possible? Same for `_after-install_` it sounds like it should be `_encryption.after_` ... same thought for all the others that are called `_before` or `_after` ... anyways that is a comment for outside this PR but if you agree maybe we can create a ticket

content/en/docs/Architecture/cloud-init.md
```diff
@@ -41,32 +41,51 @@ stages:
...
+| _after-upgrade-chroot_     | This stage happens after the image has been upgraded                                                                                                                                                                                                                                |
 
 Note: Steps executed at the `chroot` stage are running inside the new OS as chroot, allowing to write persisting changes to the image, for example by downloading and installing additional software.
 
+{{% alert title="About after and before substages" color="info" %}}
+All the above stages are automatically run with a `before` and `after` suffix automatically, so users can always run something before of after the given
+stage with ease. So if a configuration is doing something in the `before-install` stage and you want to runs something before it, you could use the `before-install.before` to run it before the main stage runs.
```
mauromorales, line 74: this is exactly why I think it would make sense to rename `before-install` doesn't really tell me at which point because it means a lot of things and `before-install.before_` and `before_install.after_` are just weird imo :P

content/en/docs/Architecture/cloud-init.md
Itxaka: yes, This needs a huge overhaul to first be on sync with all actions (like there should be pre and post in all of them and have more and more different stages that we can run. like `install-partition-before` `install-partition-after` `install-encryption-before` and so on.

In other places like reset we should also have things like before oem partitioning in case you want to save something from oem, etc..

Its clear this has gotten a bit abandoned and can be much more specific, much more in line with each action and much richer in opportunities to have the users affect the outcome of them.

content/en/docs/Architecture/cloud-init.md
mauromorales: thanks for thinking so highly of me haha

and I think the problem would be fixed if we remove those `before-install` and `after-install` and just have the stages with their names and `.pre_` and `.after_` or ideally `.before_` and `.after_` suffixes.

Since that is probably a hard change I think I would go with something like this for the time being:

- [mount partitions]
- [before-install]
- [deploy active image]
- [after-install-chroot]
- ...
- [encrypt partitions]
- [after-install]
- [reboot/power off]

content/en/docs/Architecture/cloud-init.md
Itxaka: tbh I think this is wrong and I fucked up. I mean, if you run them manually via the agent they do but in here I think we just run the exact stage and nothing else, so I migth need to remove this.

content/en/docs/Architecture/cloud-init.md
jimmykarily: @mauromorales by "previous" one I meant the one above not the previous "version". Sorry for the confusion. It's a bit explained below but still the user will read these 2 in sequence and will only have to trust that we explain later and keep reading. Ideally the text shouldn't read so similar between these two.

Itxaka: Maybe in the chroot ones we can have a link to the footnote or something so we can jump directly to the comment?

jimmykarily: That would work too. A here-link with text "How is this different from `after-upgrade`?"

jimmykarily (approved):

mauromorales (approved):

PR #279, opened 2024-10-04
Bump kairos version to the upcoming one

jimmykarily:


hugo.toml
```diff
@@ -202,8 +202,8 @@ no = 'Oh snap! Sorry to hear that. Please <a href="https://github.com/kairos-io/
   url = "/docs"
 
 [[params.versions]]
-  version = "v3.1.3"
-  url = "/v3.1.3/docs"
+  version = "v3.2.1"
```
Itxaka, line 205: cant this be managed with renovate maybe?

Itxaka (approved):

hugo.toml
Itxaka: oh we already do but it needs your change to make it work better?

jimmykarily: we decided that this file has more changes that should all go in, in one PR (e.g. the k3s version), otherwise the links in docs will be broken. So we don't worry too much about automating it, because in the end, it will need manual intervention.

PR #289, opened 2024-10-16
Document how to run commands in differen shells

mauromorales:
Fixes kairos-io/kairos#2946

### DIFF
--- a/content/en/docs/Reference/configuration.md
+++ b/content/en/docs/Reference/configuration.md
@@ -855,3 +855,18 @@ stages:
 
 For each stage, a number of modules are available, that implement various useful functions.
 Read more about them in this page: [Stage modules]({{< relref "../Reference/stage_modules.md" >}})
+
+## Running commands on different shells
+
+By default, all commands are executed in the `sh` shell. However, it is possible to run commands in a different shell by prefixing the command with the executable.
+
+For example, to run a command in the `bash` shell, you can use the following syntax:
+
+```yaml
+#cloud-config
+stages:
+  boot.after:
+    - name: "do something"
+      commands:
+        - bash /path/to/script.sh
+```


Itxaka (approved):
noice

content/en/docs/Reference/configuration.md
```diff
@@ -855,3 +855,18 @@ stages:
...
+
+For example, to run a command in the `bash` shell, you can use the following syntax:
+
+```yaml
+#cloud-config
+stages:
+  boot.after:
+    - name: "do something"
+      commands:
+        - bash /path/to/script.sh
```
jimmykarily, line 871: This assumes there is a script. Maybe we should document inline commands too? e.g.

```
bash -c 'bash things here'
```

jimmykarily (approved):

content/en/docs/Reference/configuration.md
mauromorales: @jimmykarily good suggestion, I wanted to try it out and realized that if running directly in yip I don't get an issue, am I missing something? I'm running on Ubuntu so it should have the issue according to the report

config.yaml
```
#cloud-config
stages:
  default:
    - name: "Run command in bash"
      commands:
        - /home/mauro/Workspace/kairos-docs/foo.sh
        - "if [[ -f /etc/lsb-release ]]; then cat /etc/lsb-release; fi"
```

foo.sh
```
#!/bin/bash
if [[ -z $1 ]]; then
    echo "No argument supplied"
else
    echo "Argument supplied: $1"
fi
```

```
mauro@zeno:~/Workspace/kairos-docs$ yip ./config.yaml
INFO[0000] yip version v1.11.0-g87b55bb1813f1f132a2a0f8242eadd9d702ce5f2 2024-10-04 17:57:09 UTC
INFO[0000] Running stage: default
INFO[0000] Processing stage step './config.yaml.Run command in bash'. ( commands: 2, files: 0, ... )
INFO[0000] Done executing stage 'default'
```

but if I run it in Kairos, I effectively get:

```
2024-10-16T16:05:38Z INF Processing stage step 'Run command in bash'. ( commands: 1, files: 0, ... )
2024-10-16T16:05:38Z INF Command output: sh: 1: [[: not found
```

Itxaka: Interesting. Maybe our console run method is different?

6ixfalls: I think this was brought up because of my comment about the syntax not being supported in shells, but from my little research into yip it uses `sh` to run - not sure if this is overridden in Kairos, but thought would be good to know.
https://github.com/mudler/yip/blob/87b55bb1813f1f132a2a0f8242eadd9d702ce5f2/pkg/console/console.go#L52

mauromorales: For yip on the cli I use v1.11.0 https://github.com/mudler/yip/releases/tag/v1.11.0 and for kairos I see we also have this version but as a pkg https://github.com/kairos-io/kairos-agent/blob/f24511b2dd49876e05c16b33a22423df3f3b693f/go.mod#L22

If I understand correctly, they both will be executed using Go's `os/exec` so the only difference can be the underlying system where in my case yip is running on my Ubuntu full-fledged system, while in the other case the resulting Ubuntu flavor

Itxaka: yes but both use different consoles to run no? https://github.com/kairos-io/kairos-agent/blob/f24511b2dd49876e05c16b33a22423df3f3b693f/pkg/cloudinit/console.go#L47 vs https://github.com/mudler/yip/blob/87b55bb1813f1f132a2a0f8242eadd9d702ce5f2/pkg/console/console.go#L56

So maybe our runner does something weird underneath?

mauromorales: OK then I will merge this and will open a new ticket to investigate further

PR #303, opened 2024-10-30
Standardization of Styling

wcrum:
The `content/en/docs-styling.md` is a temporary development file which helps demonstrate the styling and colors of the Kairos Docs.

Within this PR, the goal will be to standardize alerts between notes, info, etc. Across different pages there is not a standard alert for info / warning / tips. 

1. Should this repository expand the type of alerts? Which ones should we standardize on?
    - https://squidfunk.github.io/mkdocs-material/reference/admonitions/#supported-types
    - Above is an example (not Hugo) of different alerts / admonitions that we could add?


mauromorales:
regarding whether we should extend them, I don't think we need it now, I think with the 3 colors we can already have most of what we want, we just need to make sure that it is clear what each color means

for me, so far red (warning) is the only clear one, this is important and within the title of the alert we can put whatever that means e.g.

```
{{% alert title="Warning" color="warning" %}}
```

or

```
{{% alert title="Breaking changes" color="warning" %}}
```

and so on

For the other ones, I'm open to suggestions, what do you think they should mean?

wcrum:
@mauromorales I agree, the warning one is the only one that is clear. I think the current note and info alerts is fine, as development notes should have prioritization over just standard information.

NIT only change I see as possibly necessary is a darker secondary alert... although not used.

For the alert blocks, what would be the standard message? Multiple Notes have the title note and first word being a Note, a bit repetitive.

mauromorales:
@kairos-io/maintainers I think this is good to merge but pinning you in case you'd like to define this before it's live

jimmykarily:
> @kairos-io/maintainers I think this is good to merge but pinning you in case you'd like to define this before it's live

Don't we need some text in that page to explain what it is about?

mauromorales:
makes sense ... @wcrum let's add a description of what this page is about and call it a day

content/en/docs/Reference/alerts.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Alerts Styling
+---
+
```
jimmykarily, line 4:
```suggestion
This page demonstrates the styling and colors of alerts in the Kairos Docs.

The goal will be to standardize alerts between notes, info, etc. across different pages.
```

content/en/docs/Reference/alerts.md
```diff
@@ -0,0 +1,24 @@
+---
+title: Alerts Styling
+---
+This page demonstrates the styling and colors of alerts in the Kairos Docs.
+
+The goal will be to standardize alerts between notes, info, etc. across different pages.
+## Alerts
```
jimmykarily, line 7:
```suggestion

## Alerts
```

jimmykarily (approved):

PR #305, opened 2024-11-07
Update meta distro requirements

mauromorales:


content/en/docs/Architecture/meta.md
```diff
@@ -21,9 +21,14 @@ Every component is extensible and modular such as it can be customized and repla
...
-- Use `grub` as bootloader
-- If the system supposedly is meant to be used with EFI, the kernel needs to have enabled the `CONFIG_EFI_STUB` option ( see: https://docs.kernel.org/admin-guide/efi-stub.html)
+- Trusted Boot Images
+  - Use a recent enough version of `systemd` as init system
+  - Use `systemd-boot` as bootloader
+- Secure Boot Only Images
+  - Either use `systemd` or `openrc` init system
+  - Use `grub` as bootloader
+- All
+  - If the system supposedly is meant to be used with EFI, the kernel needs to have enabled the `CONFIG_EFI_STUB` option ( see: https://docs.kernel.org/admin-guide/efi-stub.html)
```
jimmykarily, line 31:
```suggestion
  - If the system is meant to be used with EFI, the kernel needs to have enabled the `CONFIG_EFI_STUB` option ( see: https://docs.kernel.org/admin-guide/efi-stub.html)
```

content/en/docs/Architecture/meta.md
```diff
@@ -21,9 +21,14 @@ Every component is extensible and modular such as it can be customized and repla
 
 In order to convert a Linux Distribution to Kairos, the distribution must meet the following requirements:
 
-- Either use `systemd` or another init system
-- Use `grub` as bootloader
-- If the system supposedly is meant to be used with EFI, the kernel needs to have enabled the `CONFIG_EFI_STUB` option ( see: https://docs.kernel.org/admin-guide/efi-stub.html)
+- Trusted Boot Images
+  - Use a recent enough version of `systemd` as init system
+  - Use `systemd-boot` as bootloader
+- Secure Boot Only Images
```
jimmykarily, line 27: How does one build such an image? Do we have a flow documented that allows building and image that is signed but it's not the full blown Trusted boot one?

content/en/docs/Architecture/meta.md
```diff
@@ -21,9 +21,14 @@ Every component is extensible and modular such as it can be customized and repla
 
 In order to convert a Linux Distribution to Kairos, the distribution must meet the following requirements:
 
-- Either use `systemd` or another init system
-- Use `grub` as bootloader
-- If the system supposedly is meant to be used with EFI, the kernel needs to have enabled the `CONFIG_EFI_STUB` option ( see: https://docs.kernel.org/admin-guide/efi-stub.html)
+- Trusted Boot Images
+  - Use a recent enough version of `systemd` as init system
```
Itxaka, line 25: can we exactly specify the version here?

for Kairos 3.0.0 the minimum is systemd 252
Then for 3.1.x or 3.2.x we raised it to minimum of 256 due to bugs in the tpm stuff

content/en/docs/Architecture/meta.md
Itxaka: thats just a normal image. They all got secureboot enabled artifacts by default with the shim and such :D

jimmykarily: So they are signed with whatever key signs the shim? That changes per distro I guess?

Itxaka: Yes indeed. They just work like a normal upstream iso

content/en/docs/Architecture/meta.md
mauromorales: Very good point, I'll add it. I will only mention 256 then because now we have versioned documentation

mauromorales: Very good point, I'll add it. I will only mention 256 then because now we have versioned documentation

PR #318, opened 2024-11-26
Update configuration.md

dbaker-rh:
Update configuration reference to be more appropriate to copy/paste.


mudler (approved):
Thanks, looking good!

PR #321, opened 2024-12-03
Replace references to enki with auroraboot

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/3002


hugo.toml
```diff
@@ -111,8 +111,6 @@ k3s_version = "v1.31.2+k3s1"
 kairos_version = "v3.2.3"
 # renovate: datasource=github-releases depName=kairos-io/osbuilder
 osbuilder_version = "v0.401.0"
```
jimmykarily, line 113: we could remove this too, it's not used anywhere, but it could be used to refer to osbuilder as the controller.

mauromorales (approved):

PR #323, opened 2024-12-05
Add assesment docs

Itxaka:


jimmykarily (approved):

content/en/docs/Examples/boot_assessment_trusted_boot.md
```diff
@@ -0,0 +1,118 @@
+---
+title: "Enabling Automatic Boot Assessment with Trusted Boot"
+linkTitle: "Enabling Automatic Boot Assessment with Trusted Boot"
+weight: 4
+description: This section describe examples on how to enable Automatic Boot Assessment with Trusted Boot in your own services.
+---
+
+
+# Enabling Automatic Boot Assessment with Trusted Boot
```
mudler, line 9: I think that by having also the title here it will display it twice, not 100% sure, but we don't do it in other parts too indeed: https://raw.githubusercontent.com/kairos-io/kairos-docs/refs/heads/main/content/en/docs/Examples/k3s-stages.md

```suggestion
```

content/en/docs/Examples/boot_assessment_trusted_boot.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Step 1: Configuring a Service to Trigger `boot-complete.target`
+
+To ensure a service's success or failure impacts the boot assessment, modify its service file to interact with `boot-complete.target`:
+
+1. **Edit the Service File**:  
+
+   Override the service configuration using:
+   ```bash
+   sudo systemctl edit <service-name>
```
mudler, line 40: mmh where this commands has to be run? in a running system? how can we do that from the cloud config file?

Itxaka: incoming in a sec!

content/en/docs/Examples/boot_assessment_trusted_boot.md
```diff
@@ -0,0 +1,118 @@
...
+title: "Enabling Automatic Boot Assessment with Trusted Boot"
+linkTitle: "Enabling Automatic Boot Assessment with Trusted Boot"
+weight: 4
+description: This section describe examples on how to enable Automatic Boot Assessment with Trusted Boot in your own services.
+---
+
+
+# Enabling Automatic Boot Assessment with Trusted Boot
+
+In this tutorial, we will walk through how to configure Kairos to enable **automatic boot assessment**, where the boot loader can determine the health of a boot entry. Specifically, we'll configure systemd services to trigger the `boot-complete.target` to mark boot entries as *good* or *bad*. We'll also cover how to implement automatic reboots when a service fails, allowing retries of a boot entry until success or exhaustion of attempts.
```
mudler, line 10: By reading it entirely something is not really clear to me: here we say that the example is on how to enable the automatic boot assessment, but later we talk about marking an arbitrary success or failure impacts the boot assessment - I think we should be explicit here in answering the following questions: 

- what are system defaults ? is boot assessment enabled?
- is Kairos having a default boot assessment strategy?
-  if Kairos does have a default , how it behaves, what are the limitations? 
-  If it doesn't, why we don't have a default for it?
-  finally, how can we extend the boot assessment by including our own custom services to impact the system state check?

content/en/docs/Examples/boot_assessment_trusted_boot.md
```diff
@@ -0,0 +1,118 @@
...
+    - By default, each boot entry has **3 retries**. A failure to reach `boot-complete.target` reduces the retry count. Once retries are exhausted, another boot entry is chosen if available.
+
+3. **Service-Level Controls**:
+    - Configure services to participate in boot entry assessment and trigger reboots on failure.
+
+---
+
+## Step 1: Configuring a Service to Trigger `boot-complete.target`
+
+To ensure a service's success or failure impacts the boot assessment, modify its service file to interact with `boot-complete.target`:
```
mudler, line 37:
```suggestion
To ensure a service's failure impacts the boot assessment, modify its service file to interact with `boot-complete.target`:
```

content/en/docs/Examples/boot_assessment_trusted_boot.md
Itxaka: we currently dont have any assesment strategy...yet.

The idea was to write the docs first and then start enabling it ourselves, plus adding e2e tests for it.

Current boot assessment is set as the default, i.e. if multi-user.target is reached, then it considers the system as good and marks it as such.

But I get what you mean, I will redo this to extend to introduce first the current states of assesment on kairos and then keep this as how can be extended to support more things.

Itxaka: Expanded!

PR #330, opened 2024-12-31
fix: add the correct cncf logo

hanshal101:
FIXES: https://github.com/kairos-io/kairos/issues/2602
CLOSES: https://github.com/kairos-io/kairos/issues/2602

## Description
This PR fixes the old pixelated CNCF Logo to the official and the correct one.


Itxaka (approved):
its a bit big and the color does not match but good enough for now

PR #335, opened 2025-01-07
Add adopters section in homepage

mauromorales:
Adds a section for adopters to kairos homepage. There are three reasons why we want to add this section:

1. to allow other potential users to see what kind of companies are using Kairos 🔍 
2. to help us apply to become an incubated CNCF project ☁️ 
3. to give visibility to projects, we are proud that they are using Kairos under the hood 💪 

relates to kairos-io/devrel#1


jimmykarily:
I love it! We should get more logos there :).

Itxaka (commented):
noice!

mauromorales:
Hi @bencorrado, this is a followup from https://github.com/kairos-io/community/pull/9. Could you please review and let us know if you're ok with us adding this section, and if it's ok, then please validate that the text and logos and link are of your liking. Thanks!

bencorrado (changes requested):
Looks great, I think we could change the text to:
`DeEEP Network uses Kairos under the hood of their device to offer a highly secure system with trusted boot.`


Just adding the ly to high to make "highly".

Thanks!

PR #340, opened 2025-01-24
Update single node example docs

mauromorales:
I think this change helps to keep the single node example concise and focused on the topic without explaining other details of the k3s configuration, those should probably go on a different page.

Additionally:

- adds a real-world example of a helm chart application that can be tested
- adds the admin group to the user creation

### DIFF
--- a/content/en/docs/Examples/single-node.md
+++ b/content/en/docs/Examples/single-node.md
@@ -11,55 +11,62 @@ In the example below we will use a bare metal host to provision a Kairos node in
 
 Use the standard images which contain `k3s`.
 
-Follow the [Installation]({{< relref "../installation/" >}}) documentation, and use the following cloud config file with Kairos:
+Follow the [Installation]({{< relref "../installation/" >}}) documentation, using the following cloud config:
 
 ```yaml
 #cloud-config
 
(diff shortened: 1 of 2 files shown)


content/en/docs/Examples/single-node.md
```diff
@@ -11,55 +11,62 @@ In the example below we will use a bare metal host to provision a Kairos node in
...
-  ```
-  k3s:
-    enabled: true
-    replace_args: true
-    args:
-    - --node-name=my-node
-  ```
-  {{% alert title="Note" %}}
-  `replace_args` replaces all arguments otherwise passed to k3s by Kairos with those supplied here. Make sure you pass all the arguments you need.
-  {{% /alert %}}
```
jimmykarily, line 64: We definitely need to move these somewhere (not just delete them). I remember it was added as an explanation to someone's question on how to pass arguments to k3s. Let's make sure we keep it around.

jimmykarily (approved):

content/en/docs/Examples/single-node.md
mauromorales: Good point, I've opened to follow up https://github.com/kairos-io/kairos-docs/pull/343

PR #350, opened 2025-02-06
Replaced references to Elemental with Kairos

ig0rmin:
I noticed that in couple of places documentation mentioned Elemental where it should be Kairos.


mauromorales (approved):
thanks!

PR #353, opened 2025-02-17
Call hightlightAll of Prism manually

jimmykarily:
after we have replaced variables.

Fixes: https://github.com/kairos-io/kairos/issues/3159


mauromorales (approved):
cool 😎

PR #357, opened 2025-02-20
Add dark mode :sunglasses:

mauromorales:
Because otherwise whenever I'm checking references late at night without lights in the room then I burn my eyes 🤣


Itxaka:
dark mode on the main page just causes the Downloads button to be terrible xD

image

Itxaka (approved):
other than the small Downloads issue, rest looks much nicer in dark mode, especially the code snippets, nice!

layouts/partials/hooks/head-end.html
```diff
@@ -1,10 +1,4 @@
-<script>
-        // https://prismjs.com/docs/Prism.html#.manual
-        // We want to call replaceVariables before we highlight code
-        window.Prism = window.Prism || {};
-        Prism.manual = true;
-</script>
```
jimmykarily, line 6: this was the fix for this: https://github.com/kairos-io/kairos/issues/3159

I see it still works but I don't see why./

mauromorales: I completely removed prism because there was no need for it, we don't have too many languages that need to be highlighted, chroma which is the default in docsy can do a good job with our bash and yaml for the most part

PR #359, opened 2025-02-21
Add optional telemetry on the docs scripts

mauromorales:
In this PR:

✅  Change the flavor selection to a settings section. This will make it easier to recognize that there are more things to select besides the flavor and the release. My goal is that the user can select multiple aspects of what they are interested in (for later in the future). I think this also looks much better on the navbar

Closed
Screenshot 2025-02-21 at 13 35 43

Open
Screenshot 2025-02-21 at 13 35 30

✅ A short explanation about the telemetry can be triggered when hovered over the info button

Screenshot 2025-02-21 at 13 37 10

✅ automatically change the scripts with the different repositories with or without telemetry when the user clicks on the checkbox and saves the information on the local storage so the user gets the same experience after setting it once

✅ telemetry on by default. Yes I know this one will be polemic but I as I've stated before, I think those who find it important to remove, will configure and remove while those who don't care will never go and do an extra step for something they don't care about. Also it's fully GDPR compliant and not tracking any personal information. We can also make an announcement to not be deceiving in any way


Itxaka (approved):
we have become death, destroyers of the world

mauromorales:
@Itxaka exactly the reaction I was expecting 🤣 

@mudler @jimmykarily can I also get your review, don't want to merge this one without knowing that everyone is on-board 😅

PR #361, opened 2025-02-25
Remove unused css and add some variables

mauromorales:
While working on the new nav buttons I found all this than can be cleaned up

### DIFF
--- a/assets/scss/_variables_project.scss
+++ b/assets/scss/_variables_project.scss
@@ -6,6 +6,11 @@ Add styles or override variables from the theme here.
 
 $primary: #03153A;
 $secondary: #e1e5ee;
+$orange: #EE5007;
+$softer-orange: #FF7442;
+$softest-orange: #FB9E70;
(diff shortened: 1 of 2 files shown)


assets/scss/_variables_project.scss
```diff
@@ -6,6 +6,11 @@ Add styles or override variables from the theme here.
 
 $primary: #03153A;
 $secondary: #e1e5ee;
+$orange: #EE5007;
+$softer-orange: #FF7442;
```
Itxaka, line 10: Why not orange, orangery and orangest? Wasted opportunity imho lol

Itxaka (approved):

assets/scss/_variables_project.scss
mauromorales: LOL ... I started with orange1 and orange2, then settled with this, but you have a point. I should have gone with organte and mar-a-lago-orange 🤣

PR #364, opened 2025-02-27
Update docs to refer to init and newer aurora

Itxaka:


content/en/docs/Development/development.md
```diff
@@ -10,66 +10,36 @@ Here you can find development notes intended for maintainers and guidance for ne
...
-- rsync
-
-If you are building a flavor without Earthly, be sure to consume the packages from our repository to convert it to a Kairos-based version.
+```bash
+docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v $PWD/build/:/output \
+  quay.io/kairos/auroraboot:v0.5.0 build-iso --output /output/ docker:myBaseKairos:v1.0.0 
+```
 
-### Bumping packages
+You can see some example of builds in either the [kairos-init repo](https://github.com/kairos-io/kairos-init/blob/main/.github/workflows/test.yml), which builds a series of OCI containers of different base images, variants and platforms or on the [Kairos repo](https://github.com/kairos-io/kairos/tree/master/.github/workflows) itself which generates not only different types of platforms and variants and base images but also generates different types of artifacts, like Isos, Trusted Boot isos, Trusted Boot upgrade artifacts, Raw Disk images and so on.
```
mauromorales, line 40:
```suggestion
You can see some example of builds in either the [kairos-init repo](https://github.com/kairos-io/kairos-init/blob/main/.github/workflows/test.yml), which builds a series of OCI containers of different base images, variants and platforms or on the [Kairos repo](https://github.com/kairos-io/kairos/tree/master/.github/workflows) itself which generates not only different types of platforms and variants and base images but also generates different types of artifacts, like ISOs, Trusted Boot ISOs, Trusted Boot upgrade artifacts, raw disk images and so on.
```

content/en/docs/Getting started/what-is-kairos.md
```diff
@@ -115,18 +115,21 @@ git clone https://github.com/kairos-io/kairos.git
 cd kairos
 ```
 
+Then we can build an OCI artifacts which is the base of all things Kairos. Our ISOs, cloud images, upgrade artifacts are all OCI artifacts and they are managed like a normal OCI artifact. Can be pushed, tagged, labelled, build on top by using it as FROM in a Dockerfile, etc..
```
mauromorales, line 118:
```suggestion
Then we can build an OCI artifact which is the base of all things Kairos. Our ISOs, cloud images, upgrade artifacts are all OCI artifacts and they are managed like a normal OCI artifact. They can be pushed, tagged, labelled, build on top by using it as FROM in a Dockerfile, etc..
```

content/en/docs/Reference/kairos-factory.md
```diff
@@ -17,75 +17,357 @@ For the newcomer or someone who simply needs an immutable OS with k3s and edgeVP
...
+
+{{% /alert %}}
+
+## Phases
+
+kairos-init is divided in 2 phases, one its the install phase which install all nedeed packages and binaries and the other is the init phase, which gets the system ready. This are the main parts of each phase:
+
+Install:
+ - Install required packages via system package manager
+ - Install [kairos-framework]() into the system
```
mauromorales, line 185:
```suggestion
 - Install [kairos-framework](https://github.com/kairos-io/kairos-framework) into the system
```

content/en/docs/Reference/kairos-factory.md
```diff
@@ -17,75 +17,357 @@ For the newcomer or someone who simply needs an immutable OS with k3s and edgeVP
...
+ - Remove all existing initrds
+ - Build a new initrd and link it to `/boot/initrd`
+ - Enable/disable needed services
+ - Run some workarounds (grub vs grub2 naming for example)
+ - Run system cleanup to avoid leftovers
+
+
+As you can see, both of this stages runs separatedly so you can hijack this in the middle and add or remove things.
+
+For example, its possible to add extra modules to be added to the initrd, or a specific kernel instead of the default latest one.
```
mauromorales, line 201:
```suggestion
For example, it's possible to add extra modules to be added to the initrd, or a specific kernel instead of the default latest one.
```

content/en/docs/Reference/kairos-factory.md
```diff
@@ -17,75 +17,357 @@ For the newcomer or someone who simply needs an immutable OS with k3s and edgeVP
...
+Extra services can be either added or made required, etc..
+
+The separation its also very useful for caching, as once the install phase has been cached by docker we can modify the following steps to fix any issues before the init phase is run, without removing the cache, which is really good under cross platforms like aarch64 where installing packages is very slow unless you are on the native platform.
```
mauromorales, line 204:
```suggestion
The separation it's also very useful for caching, as once the install phase has been cached by docker we can modify the following steps to fix any issues before the init phase is run, without removing the cache. This is particularly useful under cross platform builds where speed can take a big impact depending on your setup. 
```

content/en/docs/Installation/raspberry.md
```diff
@@ -32,7 +32,7 @@ Extract the `img` file from a container image as described [in this page]({{< re
 Plug the SD card to your system. To flash the image, you can either use Etcher or `dd`. Note it's compressed with "XZ", so we need to decompress it first:
 
 ```bash {class="only-flavors=openSUSE+Leap-15.6,openSUSE+Tumbleweed,Ubuntu+20.04,Ubuntu+22.04,Alpine+3.19"}
-xzcat {{<image variant="standard" model="rpi4" arch="arm64" suffix=".img.xz">}} | sudo dd of=<device> oflag=sync status=progress bs=10MB
+sudo dd if={{<image variant="standard" model="rpi4" arch="arm64" suffix=".img.xz">}} of=<device> oflag=sync status=progress bs=10MB
```
jimmykarily, line 35: if the suffix is `.img.xz` then `xzcat` is still needed.

content/en/docs/Reference/auroraboot.md
```diff
@@ -599,62 +603,58 @@ This method differs from the ones documented in the [Build raw images with QEMU]
...
   --set "state_dir=/aurora"
 ```
 
-The raw disk image will be available in `iso/disk.raw`
+The raw disk image will be available in the current directory with the `.raw` image extension
+
+To generate a BIOS raw image just change `disk.efi=true` for `disk.bios=true`
 
-The cloud config file (`config.yaml`) should look like the following:
+The cloud config file (`01_reset.yaml`) in the raw image oem partition should look like the following:
```
jimmykarily, line 620: Is this manual step really needed? Don't we do this by default?

content/en/docs/Reference/image_matrix.md
```diff
@@ -60,14 +60,21 @@ The artifacts can be found in the `build` directory.
 
 ### Building core and standard generic images
 
-Unfortunately we don't have the resources and capacity to build every possible artifact in our matrix. Thankfully, you can still build those images manually on your local machine, all you need is [git](https://git-scm.com/), [docker](https://www.docker.com/) and [earthly](https://earthly.dev/). Here's an example how to build an Almalinux ARM generic image
+Unfortunately we don't have the resources and capacity to build every possible artifact in our matrix. Thankfully, you can still build those images manually on your local machine, all you need is [git](https://git-scm.com/), [docker](https://www.docker.com/). Here's an example how to build an Almalinux ARM RPI4 container image.
```
jimmykarily, line 63:
```suggestion
Unfortunately we don't have the resources and capacity to build every possible artifact in our matrix. Thankfully, you can still build those images manually on your local machine, all you need is [git](https://git-scm.com/) and [docker](https://www.docker.com/). Here's an example how to build an Almalinux ARM RPI4 container image.
```

content/en/docs/Reference/image_matrix.md
```diff
@@ -60,14 +60,21 @@ The artifacts can be found in the `build` directory.
...
 
 ```bash
 git checkout https://github.com/kairos-io/kairos.git
 cd kairos
-earthly +all-arm-generic --FAMILY=@family --FLAVOR=@flavor --FLAVOR_RELEASE=@flavorRelease --BASE_IMAGE=@baseImage --VARIANT=core
+docker build --platform arm64 --build-arg BASE_IMG=almalinux:9 --build-arg MODEL=rpi4 --build-arg VERSION=1.0.0 -f images/Dockerfile -t mycustomimage:1.0.0 .
```
jimmykarily, line 68: Is `--platform linux/arm64` the same as `--platform arm64`? I've always used the first one.

content/en/docs/Reference/kairos-factory.md
```diff
@@ -17,75 +17,353 @@ For the newcomer or someone who simply needs an immutable OS with k3s and edgeVP
...
+| -s           | Sets the stage to run                                    | install, init, all          | all             |
+| -m           | Sets the model                                           | generic, rpi3, rpi4         | generic         |
+| -v           | Sets the variant                                         | core, standard              | core            |
+| -k           | Sets the Kubernetes provider                             | k3s,k0s                     | k3s             |
+| -r           | Sets registry and org (Checks this repo when upgrading)  | Any valid registry and org  | quay.io/kairos  |
+| -t           | Sets Trusted Boot on                                     | true,false                  | false           |
+| -f           | Sets the framework version to use                        | Any valid framework version | latest          |
+| --fips       | Use FIPS framework for FIPS 140-2 compliance images      | true,false                  | false           |
+| --k8sversion | Sets the provider version to use                         | Any valid provider version  | latest          |
```
jimmykarily, line 141: can't we already say: k3s or k0s ?

content/en/docs/Reference/kairos-factory.md
```diff
@@ -17,75 +17,353 @@ For the newcomer or someone who simply needs an immutable OS with k3s and edgeVP
...
+ - Remove Old kernel links and duplicates
+ - Get and link kernel to `/boot/vmlinuz`
+ - Remove all existing initrds
+ - Build a new initrd and link it to `/boot/initrd`
+ - Enable/disable needed services
+ - Run some workarounds (grub vs grub2 naming for example)
+ - Run system cleanup to avoid leftovers
+
+
+As you can see, both of these stages runs separatedly so you can hijack this in the middle and add or remove things.
```
jimmykarily, line 195:
```suggestion
As you can see, both of these stages runs separately so you can hijack this in the middle and add or remove things.
```

content/en/docs/Reference/kairos-factory.md
```diff
@@ -17,75 +17,353 @@ For the newcomer or someone who simply needs an immutable OS with k3s and edgeVP
...
 ```bash
-earthly +base-image \
-  --BASE_IMAGE=my-base-image \ # The image we just built
-  --FAMILY=ubuntu \ # Kairos groups images by family, in this case Ubuntu has its own
-  --FLAVOR=ubuntu \ # The flavor of the image, in this case Ubuntu
-  --FLAVOR_RELEASE=24.04 \ # The release of the flavor
-  --MODEL=generic \ # We are building for a generic model which can be used on a VM or hardware but not on a RPi or similar boards
-  --VARIANT=core \ # We have standard images that comes with K3s, or core images that are just the base OS
-  --BOOTLOADER=systemd-boot # The bootloader to use, in the case of UKI it has to be systemd-boot
+$ docker build --platform=arm64 -t alpine-rpi:3.19 --build-arg MODEL=rpi4 --build-arg BASE_IMAGE=alpine:3.19 --build-arg VERSION=v1.0.0 .
```
jimmykarily, line 232: does Alpine work? :p

content/en/docs/Installation/raspberry.md
Itxaka: ah crap, this is wrong, is not the proper suffix, good catch!

content/en/docs/Reference/auroraboot.md
Itxaka: yes, but its more of an informative thing rigth? before we mentioned it so maybe now we also need to do so? no idea, maybe this needs a rework

content/en/docs/Reference/image_matrix.md
Itxaka: if no OS is given it defaults to the current. So yeah usually the same unless you run from mac in which case it will be darwin/arm64 lol. Let me change this

content/en/docs/Reference/kairos-factory.md
Itxaka: well, those are providers no? So any valid provider version is correct and it opens to have more providers (rke2 maybe?) so I think this is correct and concise.

content/en/docs/Reference/kairos-factory.md
Itxaka: it will.

PR #366, opened 2025-02-28
Add documentation about Google Cloud

jimmykarily:
Part of: https://github.com/kairos-io/kairos/issues/3147

Needs: https://github.com/kairos-io/kairos/pull/3213

### DIFF
--- a/content/en/docs/Installation/gce.md
+++ b/content/en/docs/Installation/gce.md
@@ -12,7 +12,7 @@ If you want to build a custom image, you can follow the instructions in the [Bui
 ## Prerequisites
 
 - A Google Cloud account with permissions to create VMs.
-- A Google Cloud compatible image of Kairos. You can use the public image provided by Kairos or build your own image.
+- A Google Cloud compatible image of Kairos. You can use the public image provided by Kairos or [build your own image]({{< ref "/docs/Reference/auroraboot.md#generate-raw-disk-images" >}}).
 
 ## Deploy a VM
 
--- a/content/en/docs/Installation/gce.md
+++ b/content/en/docs/Installation/gce.md
@@ -21,6 +21,10 @@ Unfortunately Google Cloud [doesn't allow users to search among public images in
 1. Make sure you are authenticated with the cli: `gcloud auth login`
 1. Create a VM using the latest Kairos image:
 
+{{% alert title="Note" color="warning" %}}
+As described below, it is possible to reset to any desired image on first boot. That's the reason only one Kairos flavor is published in Google Cloud (Ubuntu 24.04). This allows us to save costs and time by not pushing unnecessary artifacts.
+{{% /alert %}}
+
 
 ```bash {class="only-flavors=Ubuntu+24.04"}
 gcloud --project  <your_project_here> compute instances create kairos-vm-test \
--- a/content/en/docs/Installation/gce.md
+++ b/content/en/docs/Installation/gce.md
@@ -12,7 +12,7 @@ If you want to build a custom image, you can follow the instructions in the [Bui
 ## Prerequisites
 
 - A Google Cloud account with permissions to create VMs.
-- A Google Cloud compatible image of Kairos. You can use the public image provided by Kairos or [build your own image]({{< ref "/docs/Reference/auroraboot.md#generate-raw-disk-images" >}}).
+- A Google Cloud compatible image of Kairos. You can use the public image provided by Kairos (see below) or [build your own image]({{< ref "/docs/Reference/auroraboot.md#generate-raw-disk-images" >}}) and upload it to your google project ([check how the Kairos CI does it](https://github.com/kairos-io/kairos/blob/48d5c2bc8fc5555263f799db8a3388d7d46cd559/.github/workflows/upload-cloud-images.yaml#L36-L89)).
 
 ## Deploy a VM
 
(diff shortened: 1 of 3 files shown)


content/en/docs/Installation/gce.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Deploy a VM
+
+Unfortunately Google Cloud [doesn't allow users to search among public images in different projects](https://cloud.google.com/compute/docs/images/managing-access-custom-images#share-images-publicly). One has to know the exact project and name of the image they intend to use, even if its publicly accessible to [all authenticated users](https://cloud.google.com/compute/docs/images/managing-access-custom-images#limitations). Using the public image seems to only be possible through the command line because of the above.
+
+1. Make sure you are authenticated with the cli: `gcloud auth login`
+1. Create a VM using the latest Kairos image:
+
+
+_TODO: templetize the image_
```
jimmykarily, line 25: TODO

hugo.toml
```diff
@@ -105,6 +105,7 @@ provider_version = "v2.9.2"
 # renovate: datasource=github-releases depName=k3s-io/k3s
 k3s_version = "v1.32.1+k3s1"
 kairos_version = "master"
+google_cloud_image_version = "v3-2-1"
```
jimmykarily, line 108: I can't think of a better way to set this without making it too complex. Google cloud doesn't allow dots `.` in the name of the image and it doesn't allow users to search for our images (only certain public images are searchable, not all publicly accessible images). So we need to keep this up to date but we don't keep all versions around so we can't base this on the `kairos_version` above (e.g. we don't push `master` and we clean up older versions).

Itxaka: Isn't there a way of maybe making renovate pull the latest Kairos tag and transform it into the gcloud valid version?

Itxaka (approved):

hugo.toml
jimmykarily: I'm a noob renovate user. This is some serious black belt stuff you are discussing here :D

jimmykarily: chatgpt suggests something like this:

```
{
  "regexManagers": [
    {
      "fileMatch": ["your-file\\.ext"],  // Replace with your actual file (e.g., "config.yaml")
      "matchStrings": ["your-dependency: 'v(?<version>\\d+\\.\\d+\\.\\d+)'"],
      "replaceString": "your-dependency: 'v${version//./-}'",
      "datasourceTemplate": "npm"
    }
  ]
}
```

looking at the docs, it may be possible. But that's what I had in mind when I said `without making it too complex` :D . I can give it a try but we won't know if it works  until we have a new release. That's a slow feedback loop.

Itxaka: You can have it nice or you can have it easy, not both 🤣

When I test this renovate stuff I just put an older version in place and then run renovate locally with dry run to see if it would have update it and to which version.

Also @mauromorales is the renovate master maybe he can tell us if that would work?

mauromorales: I feel like chatgpt is BSing there, I don't see any replaceString in the docs

layouts/shortcodes/googleVersion.html
```diff
@@ -0,0 +1 @@
+{{- $.Page.Site.Params.softwareVersions.google_cloud_image_version -}}
```
mauromorales, line 1: I think what you want is just take the `kairos_version` and transform it to your liking here with go instead of having your own variable

mauromorales:
```suggestion
{{- $.Page.Site.Params.softwareVersions.kairos_version | replace "." "-" -}}
```

content/en/docs/Installation/gce.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Deploy a VM
+
+Unfortunately Google Cloud [doesn't allow users to search among public images in different projects](https://cloud.google.com/compute/docs/images/managing-access-custom-images#share-images-publicly). One has to know the exact project and name of the image they intend to use, even if its publicly accessible to [all authenticated users](https://cloud.google.com/compute/docs/images/managing-access-custom-images#limitations). Using the public image seems to only be possible through the command line because of the above.
+
+1. Make sure you are authenticated with the cli: `gcloud auth login`
+1. Create a VM using the latest Kairos image:
+
+
+```
```
mauromorales, line 25: If this only works with Ubuntu 24.04 then you can add the only-flavors here ... if it's with multiple flavors then I guess there should also be a shortcode for that?

```suggestion
```bash {class="only-flavors=Ubuntu+24.04"}
```

layouts/shortcodes/googleVersion.html
jimmykarily: this will use `master` when user is browsing the `dev` branch no? We don't push master images to google cloud :(

content/en/docs/Installation/gce.md
jimmykarily: We only push ubuntu 24.04, hardcoded: https://github.com/kairos-io/kairos/blob/48d5c2bc8fc5555263f799db8a3388d7d46cd559/.github/workflows/upload-cloud-images.yaml#L65

It doesn't matter because people are supposed to reset to whatever image they want.

jimmykarily: I commited the suggestion but I'm not sure it's right. We don't want to appear as if only ubuntu  is supported on public cloud, because that's not true. It's just that we only push one image because that can be used to install any image. The choice of flavor we push was random.

layouts/shortcodes/googleVersion.html
mauromorales: @jimmykarily indeed but imo that is ok because at the end of the day, no one should be viewing dev branch unless they know what they are doing in which sense they'd know we don't publish those

mauromorales: I guess I coudl also add some functionality so we add an automatic alert on top of those sources that we don't publish on master. I'll create a ticket

mauromorales: https://github.com/kairos-io/kairos/issues/3234

jimmykarily: switching to an older version of the docs will point to older image which we may have cleaned up by that point (we only keep 5 versions around). We don't show very old versions in docs though so this might not be a problem either (or is it?).

content/en/docs/Installation/gce.md
jimmykarily: done

hugo.toml
jimmykarily: changed the logic, using kairos_version now.

PR #368, opened 2025-03-03
Point to latest stable as default

mauromorales:


mauromorales:
not exactly what I need

PR #369, opened 2025-03-03
Display a banner on development docs

mauromorales:
So users can know that some things might not match with the stsable release


hugo.toml
```diff
@@ -96,13 +96,15 @@ github_branch = "main"
...
+development = true
+
 [params.softwareVersions]
 # renovate: datasource=github-releases depName=kairos-io/provider-kairos
 provider_version = "v2.9.2"
 # renovate: datasource=github-releases depName=k3s-io/k3s
 k3s_version = "v1.32.1+k3s1"
-# renovate: datasource=github-releases depName=kairos-io/kairos
-kairos_version = "v3.3.1"
+kairos_version = "master"
```
mauromorales, line 107: otherwise the "dev" documentation, points to a stable release which might be confusing

jimmykarily (approved):

PR #382, opened 2025-03-26
update single-node cluster example

mauromorales:
without an EdgeVPN ip, the p2p single-node will refuse to configure itself, hence the need for a clarification here

### DIFF
--- a/content/en/docs/Examples/single-node-p2p.md
+++ b/content/en/docs/Examples/single-node-p2p.md
@@ -9,6 +9,10 @@ This feature is experimental and has only been tested on local setups. Run in pr
 Feedback and bug reports are welcome, as we are improving the p2p aspects of Kairos.
 {{% /alert %}}
 
+{{% alert title="Join an EdgeVPN network" color="warning" %}}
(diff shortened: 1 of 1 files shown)


content/en/docs/Examples/single-node-p2p.md
```diff
@@ -9,6 +9,10 @@ This feature is experimental and has only been tested on local setups. Run in pr
 Feedback and bug reports are welcome, as we are improving the p2p aspects of Kairos.
 {{% /alert %}}
 
+{{% alert title="Join an EdgeVPN network" color="warning" %}}
+Make sure that you join an existing EdgeVPN network with other machines (not necessarily Kubernetes nodes) otherwise the single-node cluster will never get an EdgeVPN IP and therefore won't be configurable.
+{{% /alert %}}
```
jimmykarily, line 14: how so? They only get IP address when there is more than one in the network? Any links supporting this?

mauromorales: @jimmykarily good question, I don't have any reference to it, @mudler do you know if there's docs or code that we can point to?

PR #385, opened 2025-03-28
Update multi node ha to remove alert for k0s

mauromorales:
K0s multi node ha is supported once this is merged https://github.com/kairos-io/provider-kairos/pull/710


Itxaka (approved):

mauromorales:
looking into the configuration further I'm not sure if this still makes sense, will review later

PR #386, opened 2025-03-28
Add manual ha configuration example for k0s

mauromorales:
Also renames the examples so they are grouped together


mauromorales:
need to validate if this makes sense according to the docs

mauromorales:
decided I'll add some alerts but will not block merging this for now

mauromorales:
fixes https://github.com/kairos-io/kairos/issues/3127

Itxaka (approved):
noice!

PR #392, opened 2025-04-16
Create a page to consolidate cloud image creation

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/3208


content/en/docs/Reference/auroraboot.md
```diff
@@ -386,6 +389,7 @@ cloud_config: |
 | `cloud_config`        | Cloud config path to use for the machines. A URL can be specified, use `-` to pass-by the cloud-config from _STDIN_                                                                                                                                               |
 | `iso.data`            | Defines a path to be embedded into the resulting iso. When booting, the files will be accessible at `/run/initramfs/live`                                                                                                                                         |
 | `netboot.cmdline`     | Override the automatically generated cmdline with a custom one to use during netboot. `config_url` and `rootfs`  are automatically constructed. A reasonable value can be `netboot.cmdline=rd.neednet=1 ip=dhcp rd.cos.disable netboot install-mode console=tty0` |
+| `disk.state_size`     | Set the minimum size (in MB) for the state partition in raw disk images. By default, the state partition is sized to 3 times the size of the current image plus some additional space for system files. Use this option to override the default size if you need to accommodate larger future images. |
```
Itxaka, line 392: can we add `disk.size` here as well? It marks the final size image that will be generated. By default it uses the smallest possible size based on the actual data, but maybe its useful to let people now that if you set `disk.size=16000` you will get a final image of 16Gb

content/en/docs/Reference/auroraboot.md
```diff
@@ -615,8 +619,21 @@ docker run --privileged -v /var/run/docker.sock:/var/run/docker.sock \
   --set "state_dir=/aurora"
 ```
 
-The raw disk image will be available in the current directory with the `.raw` image extension
+The raw disk image will be available in the current directory with the `.raw` image extension. By default, the state partition will be sized to 3 times the size of the current image plus some additional space for system files.
 
+If you need to ensure the state partition can accommodate larger future images, you can override the default size with `disk.state_size`:
+
+```bash
+docker run --privileged -v /var/run/docker.sock:/var/run/docker.sock \
```
Itxaka, line 627: `--privileged` or `-P` :P

content/en/docs/Reference/auroraboot.md
Itxaka: there is also no mention of `disk.efi` or `disk.bios` here, should we add it?

content/en/docs/advanced/creating_custom_cloud_images.md
```diff
@@ -0,0 +1,151 @@
...
+
+The first step is to create a base container image using [`kairos-init`](https://github.com/kairos-io/kairos-init). This tool allows you to create a custom base image from popular distributions.
+
+Here's a basic example of creating a base image:
+
+```bash
+docker build -t my-custom-image - <<EOF
+ARG BASE_IMAGE=ubuntu:24.04
+
+FROM quay.io/kairos/kairos-init:v0.4.3 AS kairos-init
```
Itxaka, line 36: 0.4.3 or maybe even better lets take this out into a var thats gets updated by renovate like others around?

content/en/docs/advanced/creating_custom_cloud_images.md
```diff
@@ -0,0 +1,151 @@
...
+See the [Customizing Images]({{< ref "/docs/advanced/customizing" >}}) guide for detailed instructions.
+
+## Step 3: Building Bootable Images with AuroraBoot
+
+After creating and customizing your base image, you can use AuroraBoot to create bootable images:
+
+```bash
+docker run -v "$PWD"/build:/tmp/auroraboot \
+             -v /var/run/docker.sock:/var/run/docker.sock \
+             --rm -ti quay.io/kairos/auroraboot \
```
Itxaka, line 73: no version will default to latest, do we want that?

content/en/docs/advanced/creating_custom_cloud_images.md
```diff
@@ -0,0 +1,151 @@
...
+             --set "state_dir=/tmp/auroraboot"
+```
+
+{{% alert title="Note" color="info" %}}
+For more details about AuroraBoot options and configurations, see the [AuroraBoot documentation]({{< ref "/docs/reference/auroraboot" >}}).
+{{% /alert %}}
+
+### State Partition Sizing
+
+When creating cloud images, it's important to consider the size of the state partition. The state partition is created at image build time and cannot be resized later. This partition needs to accommodate all images (passive, active, and recovery) that might be used with the system.
```
Itxaka, line 87: state only stores active+passive and a transition image for upgrades, recovery is not in this partition

content/en/docs/advanced/creating_custom_cloud_images.md
```diff
@@ -0,0 +1,151 @@
...
+             --rm -ti quay.io/kairos/auroraboot \
+             --set container_image=my-custom-image \
+             --set "disable_http_server=true" \
+             --set "disable_netboot=true" \
+             --set "disk.state_size=6000" \
+             --cloud-config /path/to/cloud-config.yaml \
+             --set "state_dir=/tmp/auroraboot"
+```
+
+The value is specified in megabytes (MB). When setting a custom size, make sure it's at least 3 times the size of the largest image you plan to use, plus some additional space for system files. This ensures there's enough space for the passive, active, and recovery images, plus some overhead for future updates.
```
Itxaka, line 103: ditto

Itxaka (changes requested):
couple of changes

content/en/docs/Reference/auroraboot.md
jimmykarily: added disk.{efi|size|bios|vhd|gce}

content/en/docs/advanced/creating_custom_cloud_images.md
jimmykarily: good point. Fixed

content/en/docs/advanced/creating_custom_cloud_images.md
jimmykarily: Switched to use existing variable.

Itxaka (approved):

mudler (approved):

PR #396, opened 2025-04-25
Clarify the usage of versioning

Itxaka:
Part of https://github.com/kairos-io/kairos/issues/3363

### DIFF
--- a/content/en/docs/Reference/kairos-factory.md
+++ b/content/en/docs/Reference/kairos-factory.md
@@ -123,7 +123,7 @@ When using kairos-init, the --version argument you set isn't just cosmetic — i
 
  - Tracking changes: It helps users, automation, and debugging processes know exactly what version of a system they are running.
 
-  - Compatibility validation: Different components, like trusted boot artifacts or upgrade servers, rely on accurate versioning to operate properly.
+ - Compatibility validation: Different components, like trusted boot artifacts or upgrade servers, rely on accurate versioning to operate properly.
 
 {{% alert title="Important!" color="warning" %}}
 Kairos Factory prepares base artifacts. It’s the responsibility of the derivative project or user (you!) to define, manage, and increment the versioning of your images.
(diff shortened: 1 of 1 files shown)


content/en/docs/Reference/kairos-factory.md
```diff
@@ -114,6 +114,40 @@ config.yaml  kairos.iso  kairos.iso.sha256  netboot  temp-rootfs
...
+
+## Why the Version Matters
+
+When using kairos-init, the --version argument you set isn't just cosmetic — it defines the version metadata for the image you're building. This version is embedded into /etc/kairos-release inside the image, and it becomes critical for:
+
+ - Upgrade management: Kairos upgrade tooling checks versions to decide when and how to upgrade systems safely.
+
+ - Tracking changes: It helps users, automation, and debugging processes know exactly what version of a system they are running.
+
+  - Compatibility validation: Different components, like trusted boot artifacts or upgrade servers, rely on accurate versioning to operate properly.
```
sdwilsh, line 126: nit: extra space here in the indentation

sdwilsh (approved):
This is way more clear; thanks!

content/en/docs/Reference/kairos-factory.md
Itxaka, line 126:
```suggestion
 - Compatibility validation: Different components, like trusted boot artifacts or upgrade servers, rely on accurate versioning to operate properly.
```

jimmykarily:
> > This document makes versioning sound important and warns about potential issues that might arise from improper versioning but it's not clear what kind of verioning (bad) practices will cause issues with ugprades and the rest. For example it says we are not enforcing semver but where is the limit on what works and what doesn't?
> 
> isnt this part:
> 
> ```
> but Semantic Versioning (semver.org) conventions are expected so upgrades and compatibility checks work predictably.
> ```
> 
> Already kind of cover that? Like if you do wrong versioning (like random numbers) then upgrades migth not work properly and such? Or what did you have in mind exactly?

As we discussed in the planning call, I was mostly confused about this part:

```
Kairos does not impose a versioning cadence or scheme
```
since the rest of the sentence (and document), implies that we require the version to follow semver (thus we impose a semver scheme).

PR #408, opened 2025-05-22
Extract quick start

mauromorales:
I want the Quick start to be less prone to break, this is the first thing a newcomer will be exposed to and having broken download links and such is not a good first impression. It should work more like the blog than like the docs so to speak. Therefore, I've extracted the quick start to a section of it's own.


ci-robbot (commented):
Overall, the PR is well-structured and the documentation is comprehensive. With the suggested improvements, it will be even more effective.

tbrasser:
Now homepage getting started link is broken

mauromorales:
@tbrasser thanks for pointing it out! should be fixed now :)

PR #409, opened 2025-05-22
Automate release

mauromorales:
What was automated:

- versions will be determined automatically based on the release/vX.Y.Z branches, it will take the latest patch of each major+minor. No need to touch hugo.toml for this anymore
- removes "static" pages from old versions, so instead of looking at /v3.1.3/blog with the old articles, links always point to /blog so you see always the latest no matter which version you are looking. Same applies to getting started and community pages


mudler (approved):
that's so meta, liking it!

PR #412, opened 2025-05-26
updates to getting started guide

mauromorales:
- use kairos-init in the build example and adapt description
- fix links

### DIFF
--- a/content/en/getting-started/building-and-upgrading.md
+++ b/content/en/getting-started/building-and-upgrading.md
@@ -27,12 +27,27 @@ As we saw in the previous section on how to configure Kairos, it is not possible
 Start by creating a file called `Dockerfile` with the following content
 
 ```
-FROM quay.io/kairos/ubuntu:24.04-standard-amd64-generic-v3.1.2-k3sv1.30.4-k3s1 
+ARG BASE_IMAGE=ruby:3.4.4-bookworm # The ruby image is based on Debian bookworm, you can do the same with any of your images as long as they use a supported distro
 
-RUN apt-get update && apt-get install -y ruby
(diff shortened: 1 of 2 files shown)


content/en/getting-started/building-and-upgrading.md
```diff
@@ -73,7 +88,7 @@ The answer to this will depend on your setup but here are two things to keep in
 
 **What if I don't want to base my image on the Kairos released artifacts?**
 
-No problem, just [build your image from scratch](/reference/build-from-scratch/)
+No problem, just [build your image from scratch](/reference/kairos-factory/)
```
Itxaka, line 91: This doesn't sync with what's above. A ove we used to say, get a published artifact then install ruby on it but now it makes it use kairos-init, thus pointing here again to Kairos init makes it weird

content/en/getting-started/building-and-upgrading.md
```diff
@@ -27,12 +27,27 @@ As we saw in the previous section on how to configure Kairos, it is not possible
 Start by creating a file called `Dockerfile` with the following content
 
 ```
-FROM quay.io/kairos/ubuntu:24.04-standard-amd64-generic-v3.1.2-k3sv1.30.4-k3s1 
-
-RUN apt-get update && apt-get install -y ruby
+ARG BASE_IMAGE=ruby:3.4.4-bookworm # The ruby image is based on Debian bookworm, you can do the same with any of your images as long as they use a supported distro
```
Itxaka, line 30: I would keep it simpler in here, especially it being the quick start. I would keep the from a published artifact and link to init below if they want a more hands on approach IMHO especially fro it to be a quick start, there is nothing faster than starting from an existing built artifact

PR #415, opened 2025-05-27
Update rpi instructions

jimmyjones2:
Updated instructions based on my experiments with a Raspberry Pi, following https://kairos.io/blog/2025/03/18/how-i-automated-my-doorbell-with-kairos-and-you-can-too/
* `COS_PERSISTENT` didn't exist in the image extracted from `quay.io/kairos/opensuse:leap-15.6-standard-arm64-rpi4-v3.3.1-k3sv1.32.1-k3s1`, I needed to add the `cloud-config` into `COS_OEM`
* Add instructions for using Auroraboot to generate image

Fixes kairos-io/kairos#3445
Fixes https://github.com/kairos-io/kairos/issues/3453


content/en/docs/Installation/raspberry.md
```diff
@@ -23,36 +23,57 @@ If you are not familiar with the process, it is suggested to follow the [quickst
...
+## Install using AuroraBoot
+
+Create `build` directory and add `cloud-config.yaml`, then run:
+
+```bash
+docker run --rm --privileged -v /var/run/docker.sock:/var/run/docker.sock \
+                --platform linux/arm64 -v $PWD/build/:/output \
+                kairos.docker.scarf.sh/kairos/auroraboot:latest \
+                --debug --set "disable_http_server=true" --set "disable_netboot=true" \
+                --set "state_dir=/output" --set "disk.raw=true" \
```
ElKiwos, line 35: I couldn't find the `disk.raw` parameter in the Auroraboot documentation or the source code. 
Shouldn't this be `disk.efi=true`?

content/en/docs/Installation/raspberry.md
```diff
@@ -23,36 +23,57 @@ If you are not familiar with the process, it is suggested to follow the [quickst
...
-$ mkdir /tmp/persistent
-$ sudo mount $PERSISTENT /tmp/persistent
-$ sudo mkdir /tmp/persistent/cloud-config
-$ sudo cp cloud-config.yaml /tmp/persistent/cloud-config
-$ sudo umount /tmp/persistent
+$ OEM=$(blkid -L COS_OEM)
+$ mkdir /tmp/oem
+$ sudo mount $OEM /tmp/oem
+$ sudo cp cloud-config.yaml /tmp/oem/90_custom.yaml
+$ sudo umount /tmp/oem
```
ElKiwos, line 76: Since the cloud-config was referenced already in the AuroraBoot command and backed into the raw image is this still needed?

content/en/docs/Installation/raspberry.md
ElKiwos: Ah I see, thank you very much. I only checked:
https://github.com/kairos-io/AuroraBoot/blob/69bccf942277af0de137764b812f17adfaaf8762/pkg/schema/config.go#L48

content/en/docs/Installation/raspberry.md
jimmyjones2: Currently there are two options - auroraboot and using the raw images. I think using the raw images is too hard (kairos-io/kairos#3453) and auroraboot is much better, so would prefer to remove this section entirely

jimmykarily: I think what @ElKiwos meant was that if the instructions were followed, [this line](https://github.com/kairos-io/kairos-docs/pull/415/files#diff-7b72672ce3fa963d45234eb91b9556c482a4258a08042ef8cd4b0609f02f4f9eR36) embeds the config in the image. This step here will be replacing that file, thus it's redundant. Maybe we should explain it a bit. E.g. "if you didn't embed the config in the image, then you can add it here like this" or something like that.

Itxaka: this is correct, one refers to creating a new image from aurora which we can bundle the config directl, while this one is explaining how to inject the config into a already made img file pushed to our repos.

Itxaka (approved):
thank you!

PR #423, opened 2025-06-10
Update check-the-signatures.md

Itxaka:
update cosign command

Fixes https://github.com/kairos-io/kairos/issues/3404

### DIFF
--- a/content/en/docs/Installation/check-the-signatures.md
+++ b/content/en/docs/Installation/check-the-signatures.md
@@ -41,7 +41,7 @@ We recommend using the latest cosign version, at the time of writing, 2.5.0
 {{% /alert %}}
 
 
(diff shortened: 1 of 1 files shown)


content/en/docs/Installation/check-the-signatures.md
```diff
@@ -36,11 +36,15 @@ $ ls
 {{<image variant="core" suffix=".iso.sha256">}}  {{<image variant="core" suffix=".iso.sha256.sig">}}
 ```
 
-Then we verify that the sha256 checksums haven't been tampered with:
+{{% alert title="Cosign version Step" color="primary" %}}
+We recommend uing the latest cosign version, at the time of writing, 2.5.0
+{{% /alert %}}
+
+
+Then we verify that the sha256 checksums haven't been tampered with (substitute $VERSION with the exact Kairos version you are verifying as teh certificate identity is the release job that signs it):
```
TheRedCyclops, line 44: Here "the" is misspelled as "teh"

PR #441, opened 2025-07-07
fix(docs): checkout should be clone

sbaeurle:
Small typo in the documentation which should be fixed to avoid confusion for copy-pasters


mauromorales (approved):
thanks!

PR #472, opened 2025-09-30
automate the release process

mauromorales:
We commonly forget to do a release in the docs, for example right now v3.5.3 is out and the website still points to v3.5.2. But worse than forgetting to update, is the fact that we don't really validate that the components versions match the release, so for example we can have documentation for Kairos v3.5.2 pointint to a kairos-init which is not the one included in that release.

### DIFF
--- a/scripts/release-automation.sh
+++ b/scripts/release-automation.sh
@@ -103,8 +103,8 @@ update_hugo_toml() {
     # Create backup
     cp hugo.toml hugo.toml.backup
     
-    # Update latest_version
-    sed -i "s/^latest_version = \".*\"/latest_version = \"$version\"/" hugo.toml
+    # Update kairos_version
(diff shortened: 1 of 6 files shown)


jimmykarily (approved):

scripts/release-automation.sh
```diff
@@ -0,0 +1,318 @@
...
+
+# Get the directory where the script is located
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+root_dir="${ROOT_DIR:-$(dirname "$SCRIPT_DIR")}"
+
+# Source the utils file
+source "${SCRIPT_DIR}/utils.sh"
+
+# Default values
+DRY_RUN=false
```
Itxaka, line 12: uhhh can we run it first with DRY_RUN true and no push? XDDD

PR #504, opened 2025-12-05
Add missing configuration fields

jasperdekeuk:
Extracted all the possible yaml fields from here: https://github.com/kairos-io/kairos-sdk/blob/60e94acea6e4470f4740db60776235304c4fcf8f/types/config/config.go#L37

This is a very quick and dirty attempt, It could probably be done much better. But it can be a start :)


mudler:
@jasperdekeuk the PR is out of sync, can you rebase and review @jimmykarily PR so we can pick it up? thanks!

PR #583, opened 2026-04-11
fix: auroraboot template delimiter

kaiehrhardt:
https://github.com/kairos-io/AuroraBoot/blob/main/internal/config/config.go#L21-L22


jimmykarily (approved):
LGTM, thanks!

PR #584, opened 2026-04-13
Move configurator from homepage to Kairos Factory page

### LINKED ISSUE #4040: Homepage feedback
- move builder to kairos factory
- add config in the kairos lab section

mauromorales:
- Removed the configurator completely, we will use something else on the user side so they can even do the building direclty on their machine
- Simplified the Download section to just be either use kairos-lab or download hadron with k3s and then link to the installation in the quick start section
- reduced the size and moved the alternatives to the right, this way we make room to show off the visual configurator later without making the page much bigger

Fixes kairos-io/kairos#4040


jimmykarily:
I look at the deploy preview but I'm not sure what I'm looking for :D .

mauromorales:
@jimmykarily I added screenshots of the change in the description of the PR

jimmykarily:
Iirc this change was requested but it seems to me that hiding this new interface under that button beats the purpose of making it easy for newcomers no? I mean, who is gonna see it there? (for one, I didn't)

mauromorales:
@jimmykarily yup we discussed yesterday on the planning that maybe it was too complex to have all of that here hence why it was moved to the Kairos Factory

@Itxaka @mudler what do you guys think?

Itxaka:
> @jimmykarily yup we discussed yesterday on the planning that maybe it was too complex to have all of that here hence why it was moved to the Kairos Factory
> 
> @Itxaka @mudler what do you guys think?

Looks good to me but I would also link from the main page , in the downloads alternatives I would add a build your ownh that links to the factory or to the image builder directly.

Also why the config in there? doesnt kairos-lab provides the config already? the less stuff int here the better, I tihkn the brew install/ curl sh is the correct step but we should NOT show the config. Let kairos-labe provide the config directly by injecting it or by serving it via anything, but lets not make it complex-er?

mauromorales:
I was talking to Ettore and we decided to remove the configurator on the website, because daedalus project will better serve this.

jimmykarily (approved):

PR #601, opened 2026-05-15
Update coco documentation

jimmykarily:
Finishing: https://github.com/kairos-io/kairos/issues/3932

### DIFF
--- a/docs/advanced/coco.md
+++ b/docs/advanced/coco.md
@@ -7,7 +7,7 @@ sidebar_position: 9
 
 :::info Confidential Containers on Kairos
 
-Kairos works with [Confidential Containers](https://confidentialcontainers.org/) on AMD SEV-SNP hardware today, using an Ubuntu-based Kairos image. Support for the Hadron-based image is on the roadmap and coming soon.
+Kairos works with [Confidential Containers](https://confidentialcontainers.org/) on AMD SEV-SNP hardware today, using various Kairos flavors — including Ubuntu and [Hadron](https://github.com/kairos-io/hadron). Hadron support is available on AWS (SEV-SNP) since [Hadron v0.3.0](https://github.com/kairos-io/hadron/releases/tag/v0.3.0).
 
 As long as the necessary virtualization and memory-encryption options are enabled on your hardware (e.g. SEV / SEV-SNP / SNP Memory Coverage in BIOS on AMD platforms, or the equivalent TDX settings on Intel), the upstream Confidential Containers project can be installed on a Kairos cluster the same way it is on any Kubernetes distribution — see the [upstream installation guide](https://confidentialcontainers.org/docs/getting-started/installation/) — and workloads then opt in by selecting the appropriate [runtime class](https://confidentialcontainers.org/docs/getting-started/workload/).
 
--- a/docs/advanced/coco.md
+++ b/docs/advanced/coco.md
@@ -5,9 +5,23 @@ sidebar_position: 9
 ---
 
 
+:::info Confidential Containers on Kairos
+
+Kairos works with [Confidential Containers](https://confidentialcontainers.org/) on AMD SEV-SNP hardware today, using an Ubuntu-based Kairos image. Support for the Hadron-based image is on the roadmap and coming soon.
+
+As long as the necessary virtualization and memory-encryption options are enabled on your hardware (e.g. SEV / SEV-SNP / SNP Memory Coverage in BIOS on AMD platforms, or the equivalent TDX settings on Intel), the upstream Confidential Containers project can be installed on a Kairos cluster the same way it is on any Kubernetes distribution — see the [upstream installation guide](https://confidentialcontainers.org/docs/getting-started/installation/) — and workloads then opt in by selecting the appropriate [runtime class](https://confidentialcontainers.org/docs/getting-started/workload/).
+
+Because Kairos no longer ships prebuilt Ubuntu images by default, you'll want to build your own Kairos image with the kernel command-line flags and any vendor packages (Intel TDX/SGX attestation libraries, container image decryption tooling, etc.) that your scenario needs. See:
+
+- [Bring Your Own Image (BYOI)](/docs/reference/byoi/) — the recommended path for Ubuntu and other non-Hadron bases.
+- [The Kairos Factory](/docs/reference/kairos-factory/) — how to turn a customized base image into a Kairos-ready artifact.
+
+:::
+
+
 :::warning Warning
 
-This page describes features that are still experimental in Kairos. There are a lot of things that can be improved and might be more streamlined in the future.
+The remainder of this page describes an earlier, experimental integration based on [`enclave-cc`](https://github.com/confidential-containers/enclave-cc) and the `coco` community bundle. It is kept for historical reference and does not reflect the current recommended approach described above.
 
 :::
 
--- a/docs/advanced/coco.md
+++ b/docs/advanced/coco.md
@@ -5,16 +5,25 @@ sidebar_position: 9
 ---
 
 
-:::info Confidential Containers on Kairos
+:::info Confidential Containers on Kairos (non-Hadron)
 
-Kairos works with [Confidential Containers](https://confidentialcontainers.org/) on AMD SEV-SNP hardware today, using various Kairos flavors — including Ubuntu and [Hadron](https://github.com/kairos-io/hadron). Hadron support is available on AWS (SEV-SNP) since [Hadron v0.3.0](https://github.com/kairos-io/hadron/releases/tag/v0.3.0).
+Kairos works with [Confidential Containers](https://confidentialcontainers.org/) on AMD SEV-SNP hardware today, using non-Hadron Kairos flavors (Ubuntu and other glibc-based bases).
 
-As long as the necessary virtualization and memory-encryption options are enabled on your hardware (e.g. SEV / SEV-SNP / SNP Memory Coverage in BIOS on AMD platforms, or the equivalent TDX settings on Intel), the upstream Confidential Containers project can be installed on a Kairos cluster the same way it is on any Kubernetes distribution — see the [upstream installation guide](https://confidentialcontainers.org/docs/getting-started/installation/) — and workloads then opt in by selecting the appropriate [runtime class](https://confidentialcontainers.org/docs/getting-started/workload/).
+As long as the necessary virtualization and memory-encryption options are enabled on your hardware (e.g. SEV / SEV-SNP / SNP Memory Coverage in BIOS on AMD platforms, or the equivalent TDX settings on Intel), the upstream Confidential Containers project can be installed on a Kairos cluster the same way it is on any Kubernetes distribution (see the [upstream installation guide](https://confidentialcontainers.org/docs/getting-started/installation/)), and workloads then opt in by selecting the appropriate [runtime class](https://confidentialcontainers.org/docs/getting-started/workload/).
 
 Because Kairos no longer ships prebuilt Ubuntu images by default, you'll want to build your own Kairos image with the kernel command-line flags and any vendor packages (Intel TDX/SGX attestation libraries, container image decryption tooling, etc.) that your scenario needs. See:
 
-- [Bring Your Own Image (BYOI)](/docs/reference/byoi/) — the recommended path for Ubuntu and other non-Hadron bases.
-- [The Kairos Factory](/docs/reference/kairos-factory/) — how to turn a customized base image into a Kairos-ready artifact.
+- [Bring Your Own Image (BYOI)](/docs/reference/byoi/): the recommended path for Ubuntu and other non-Hadron bases.
+- [The Kairos Factory](/docs/reference/kairos-factory/): how to turn a customized base image into a Kairos-ready artifact.
+
+:::
+
+
+:::warning Hadron is not yet supported
+
+[Hadron](https://github.com/kairos-io/hadron) is musl-only, and the Kata Containers host binaries shipped by the Confidential Containers operator (`kata-runtime`, `kata-monitor`, `containerd-shim-kata-v2`, `containerd-nydus-grpc`) are still dynamically linked against glibc, so they fail to `execve` on a Hadron node and the pod sandbox never comes up.
+
+Upstream Kata has added an opt-in static build ([kata-containers#13243](https://github.com/kata-containers/kata-containers/pull/13243)), and Kata 4.0.0 additionally ships a static-pie Rust shim (`runtime-rs`), but no released Kata payload yet builds the full host-binary set statically. This page will be updated once such a release is available so the standard Confidential Containers install works on Hadron unchanged.
 
 :::
 


docs/advanced/coco.md
```diff
@@ -5,9 +5,23 @@ sidebar_position: 9
 ---
 
 
+:::info Confidential Containers on Kairos
+
+Kairos works with [Confidential Containers](https://confidentialcontainers.org/) on AMD SEV-SNP hardware today, using an Ubuntu-based Kairos image. Support for the Hadron-based image is on the roadmap and coming soon.
```
jimmykarily: Hadron works now. Let's update this after 4.1.1 is out.

Itxaka (approved):

jimmykarily:
I deployed the 4.1.1 hadron image on AWS and followed our steps to create a coco workload. The `kata-as-coco-runtime` pod shows this:

```
 Validating the snapshotter-handler mapping: "qemu-coco-dev:nydus,qemu-coco-dev-runtime-rs:nydus,qemu-snp:nydus,qemu-tdx:nydus"                                                                         │
│ Using containerd drop-in files: true                                                                                                                                                                   │
│ copying kata artifacts onto host                                                                                                                                                                       │
│ WARN: Distro hadron v0.3.0 does not support TDX and the TDX related runtime classes will not work in your cluster!                                                                                     │
│ WARN: Distro hadron v0.3.0 does not support TDX and the TDX related runtime classes will not work in your cluster!                                                                                     │
│ Add Kata Containers as a supported runtime for containerd                                                                                                                                              │
│ INFO: Deploying nydus-snapshotter                                                                                                                                                                      │
│ Created symlink '/etc/systemd/system/containerd.service.requires/nydus-snapshotter.service' → '/etc/systemd/system/nydus-snapshotter.service'.                                                         │
│ Unit /etc/systemd/system/nydus-snapshotter.service is added as a dependency to a non-existent unit containerd.service.                                                                                 │
│ INFO: Configuring nydus-snapshotter                                                                                                                                                                    │
│ node/kairos labeled     
```

Not sure if it's just a false warning or really not working. 

Also the coco workload pod shows these :

```
│   Warning  FailedCreatePodSandBox  69s               kubelet            Failed to create pod sandbox: rpc error: code = Unavailable desc = failed to start sandbox "cb95f355794be8b24b63904ba3f1e5099b │
│ 871f8b1e874281b9ab7152b3c230b7": failed to create containerd container: error unpacking image: failed to prepare extraction snapshot "extract-239127193-Fgph sha256:1021ef88c7974bfff89c5a0ec4fd3160da │
│ ac6c48a075f74cff721f85dd104e68": connection error: desc = "transport: Error while dialing: dial unix:///run/nydus-snapshotter/containerd-nydus-grpc.sock: timeout"                                     │
│   Warning  FailedCreatePodSandBox  58s               kubelet            Failed to create pod sandbox: rpc error: code = Unavailable desc = failed to start sandbox "6eada378accd96034353a2b33346131881 │
│ 02cbd75dfaa0aa2c6239adb8979de4": failed to create containerd container: error unpacking image: failed to prepare extraction snapshot "extract-241676793-B0PC sha256:1021ef88c7974bfff89c5a0ec4fd3160da │
│ ac6c48a075f74cff721f85dd104e68": connection error: desc = "transport: Error while dialing: dial unix:///run/nydus-snapshotter/containerd-nydus-grpc.sock: timeout"                                     │
│   Warning  FailedCreatePodSandBox  7s (x4 over 46s)  kubelet            (combined from similar events): Failed to create pod sandbox: rpc error: code = Unavailable desc = failed to start sandbox "dc │
│ a84de2df0f16240cd8954ff0c29d015ad6e05fb10b88c3889e2c5b999a58ab": failed to create containerd container: error unpacking image: failed to prepare extraction snapshot "extract-244831260-LoIW sha256:10 │
│ 21ef88c7974bfff89c5a0ec4fd3160daac6c48a075f74cff721f85dd104e68": connection error: desc = "transport: Error while dialing: dial unix:///run/nydus-snapshotter/containerd-nydus-grpc.sock: timeout" 
```
so something is definitely not working.

jimmykarily:
Claude findings (2 of them):

# Confidential Containers on Hadron: why it fails today and what Hadron needs to change

## Summary

The Confidential Containers operator (chart 0.18.0, which deploys upstream Kata Containers and the Nydus snapshotter via kata-deploy) cannot run on a Hadron node. The reason is a libc mismatch: the Kata and Nydus binaries shipped by the operator are dynamically linked against glibc, while Hadron is a musl-only system with no glibc runtime present. The binaries therefore fail to execute at all, which cascades into the pod sandbox failures we see.

This is a Hadron-side gap. Hadron is and remains musl based, so the fix is not to add glibc to Hadron. The fix is to make the Kata and Nydus host binaries musl compatible, by shipping static or musl-linked builds instead of the stock glibc artifacts.

There is also a separate, platform-level constraint that does not belong to Hadron but determines where the Kata path can run at all (see "Platform requirement" below).

## Environment tested

- Image: `quay.io/kairos/hadron:v0.3.0-standard-amd64-generic-v4.1.1-k3s-v1.35.5-k3s1`
- OS: Hadron Linux v0.3.0, k3s v1.35.5
- Operator: confidential-containers Helm chart 0.18.0, `kata-as-coco-runtime` with `k8sDistribution: k3s`
- Kata artifacts deployed to `/opt/kata` (kata-containers VERSION 3.25.0)
- Platform: AWS SEV-SNP instance

## Observed symptom

A workload using `runtimeClassName: kata-qemu-snp` stays in `ContainerCreating` and the pod events show:

```
FailedCreatePodSandBox: failed to create containerd container: error unpacking image:
failed to prepare extraction snapshot ...: connection error:
desc = "transport: Error while dialing: dial unix:///run/nydus-snapshotter/containerd-nydus-grpc.sock: timeout"
```

The containerd configuration is correct: the `kata-qemu-snp` runtime is set to `snapshotter = "nydus"`, and `[proxy_plugins.nydus]` points at `/run/nydus-snapshotter/containerd-nydus-grpc.sock`. The socket simply never exists, because the snapshotter service is dead.

## Root cause: glibc binaries on a musl-only system

`nydus-snapshotter.service` fails on every start:

```
nydus-snapshotter.service: Failed to execute /opt/kata/nydus-snapshotter/containerd-nydus-grpc: No such file or directory
Main process exited, code=exited, status=203/EXEC
```

The binary is present (about 47 MB). The "No such file or directory" refers to the binary's ELF interpreter, not the binary itself. The kernel cannot find the dynamic loader the binary asks for, so `execve` returns ENOENT and systemd reports `203/EXEC`.

Evidence:

- The Nydus and Kata binaries request the glibc loader `/lib64/ld-linux-x86-64.so.2`.
- Hadron is musl based. It ships `/lib/ld-musl-x86_64.so.1 -> /usr/lib/libc.so` and nothing else. There is no glibc loader anywhere on the system (`find / -name 'ld-linux-x86-64.so*'` returns nothing).
- The same failure affects the core Kata runtime binaries, not just Nydus. `containerd-shim-kata-v2`, `kata-runtime`, and `kata-monitor` are all glibc linked and fail to exec for the identical reason. Only the statically linked components run (`qemu-system-x86_64`, `cloud-hypervisor`, `firecracker`, `jailer`, `nydus-overlayfs`).

So even if Nydus were patched out, the Kata shim itself could not start. The whole kata-deploy binary set is glibc dynamic.

### A loader symlink is not enough

Pointing `/lib64/ld-linux-x86-64.so.2` at the musl loader does not work. When musl's loader is forced to map these binaries it fails relocation on glibc-only symbols:

```
Error relocating .../containerd-nydus-grpc: __vfprintf_chk: symbol not found
Error relocating .../containerd-shim-kata-v2: __fprintf_chk: symbol not found
```

These are glibc fortify-source functions that musl does not implement under those names. A real glibc runtime is required, not a redirected loader.

## What Hadron needs to do

Hadron stays musl, so adding glibc to the base is not an option. The binaries themselves have to stop depending on glibc. The good news is that most of the relevant binaries are Go, and the heavy C and C++ components are already static, so this is a build and packaging problem rather than a base OS problem.

The split today:

- Glibc dynamic and failing: `containerd-shim-kata-v2`, `kata-runtime`, `kata-monitor`, `containerd-nydus-grpc`. These are CGO enabled Go binaries, which is the only reason they link glibc.
- Already static and working on musl: `qemu-system-x86_64` (including the snp and tdx variants), `cloud-hypervisor`, `firecracker`, `jailer`, `nydus-overlayfs`.
- Not relevant to the host libc: the Kata guest agent runs inside the pod VM, not on the Hadron host, so it does not need to match Hadron's libc.

Options, in recommended order:

1. Rebuild the host side Kata and Nydus Go binaries as static or musl linked (recommended).
   - Built with CGO disabled, these Go binaries become fully static and run on musl with no libc dependency at all. Where a component genuinely needs CGO (for example libseccomp in the runtime), build it statically linked against musl rather than dynamically against glibc.
   - Each of the four failing binaries needs to be confirmed buildable this way. `containerd-nydus-grpc` and `nydus-overlayfs` come from nydus-snapshotter (pure Go, static friendly). `containerd-shim-kata-v2`, `kata-runtime`, and `kata-monitor` come from kata-containers and need per component checking for CGO requirements.
   - Kairos or Hadron produces these alternate artifacts, and the Confidential Containers deployment is pointed at them instead of the stock glibc kata-deploy payload.

2. Build and maintain a musl or static kata-deploy payload image.
   - The operator pulls the Kata and Nydus binaries from a kata-deploy payload container image. Maintaining a Kairos built payload image whose binaries are static or musl linked, and overriding the operator to use it, is the packaging form of option 1. This is also the natural place to upstream the work back to kata-containers so they publish a musl or static artifact set.

A glibc compatibility shim such as gcompat is explicitly not a fit. The missing symbols above (`__vfprintf_chk` and similar) are exactly the glibc surface gcompat covers incompletely, and the Kata stack is heavy on threading and dynamic loading where such shims tend to break. Static or musl native builds avoid the problem rather than papering over it.

## Follow-up issues that will surface once the binaries can run

These are not the current blocker, but they will become visible the moment the binaries can execute, so record them now:

1. The `nydus-snapshotter.service` unit declares `RequiredBy=containerd.service` and `Before=containerd.service`. On k3s there is no `containerd.service` (containerd is embedded in `k3s.service`), so the unit is not pulled in on boot and there is a stray dependency symlink under `containerd.service.requires`. The operator logs this as "Unit ... added as a dependency to a non-existent unit containerd.service". On k3s the snapshotter start and ordering needs to be handled explicitly rather than via a containerd.service dependency.
2. Confirm the static or musl builds resolve any runtime data they need (for example NSS or locale lookups) without a glibc runtime present.

## Platform requirement: the node must be able to host VMs (the AWS SEV-SNP caveat)

This is not a Hadron bug, but it decides where the Kata path can run, so it is recorded here to prevent a predictable misunderstanding.

Confidential Containers with Kata launches one VM per pod, so the node needs KVM (`/dev/kvm`) and, for SNP attestation, SEV-SNP host capability. Two confidential computing models are easy to confuse:

- Confidential VM: the whole instance runs as an SEV-SNP encrypted VM. This is what AWS SEV-SNP instances (m6a, c6a, r6a, and similar) provide. It protects the node from the hypervisor and is fine for running Kubernetes, but the confidentiality boundary is the whole node.
- Confidential Containers: each pod runs in its own confidential micro VM, which requires the node itself to host SNP guest VMs.

An AWS SEV-SNP instance is itself a leaf SNP guest. It does not expose nested virtualization, so it cannot host the per pod VMs Kata needs. Confirmed on the instance tested: `/dev/kvm` is absent, no `svm`/`vmx` in `/proc/cpuinfo`, and `dmesg` shows `SEV: SNP running at VMPL0`. So `kata-qemu-snp` cannot boot a pod sandbox there even after the musl fix. This is OS independent: the same instance behaves identically with any distribution.

Ways to run Confidential Containers on AWS:

- Bare metal (`*.metal`) instances expose `/dev/kvm`, so the VM based Kata path works. Use `kata-qemu-coco-dev` to validate the operator without confidential hardware, or `kata-qemu-snp` for real SNP on an SNP capable metal host.
- Peer pods (the Confidential Containers cloud-api-adaptor). Instead of nesting, the operator launches a separate confidential instance per pod through the cloud API, and that instance can be an AWS SEV-SNP instance. This is the model intended for clouds without nested virtualization, and it is a different and larger deployment than the Helm chart above.

To validate the Kata path end to end, use a host that exposes `/dev/kvm`: a bare metal SEV-SNP machine, or any KVM capable host with `kata-qemu-coco-dev`. A standard, non metal cloud instance cannot run the nested VM path regardless of OS.

Note also that the kata-deploy warning `Distro hadron v0.3.0 does not support TDX` is benign here. The hardware is AMD SEV-SNP, not Intel TDX, and it only disables the unused TDX runtime classes.

## Suggested acceptance criteria for the Hadron change

1. `containerd-shim-kata-v2`, `kata-runtime`, `kata-monitor`, and `containerd-nydus-grpc` all execute on Hadron (for example `containerd-nydus-grpc --version` returns successfully).
2. `nydus-snapshotter.service` reaches active state and creates `/run/nydus-snapshotter/containerd-nydus-grpc.sock`.
3. On a host with KVM and SEV-SNP support, a `kata-qemu-snp` pod reaches `Running`.

jimmykarily:
The Nydus binary kata-deploy installs (containerd-nydus-grpc) is glibc-linked, so on Hadron (musl) it fails to exec, nydus-snapshotter.service dies with 203/EXEC, and pods fail with dial
.../containerd-nydus-grpc.sock: timeout. No glibc and no rebuild needed: nydus-snapshotter publishes a fully static (CGO_ENABLED=0, -extldflags -static) binary per release.

For the version we install (v0.15.10):
- https://github.com/containerd/nydus-snapshotter/releases/tag/v0.15.10
- https://github.com/containerd/nydus-snapshotter/releases/download/v0.15.10/nydus-snapshotter-v0.15.10-linux-static.tar.gz

Verified on a Hadron node: dropping that static containerd-nydus-grpc in place makes the service go active and the socket appear.

To consume it the supported way (the binaries are baked into the kata-deploy image, so there is no per-binary flag): build a kata-deploy image with the static nydus binary swapped in and point the
chart at it:

valuesContent: |-
  kata-as-coco-runtime:
    k8sDistribution: k3s
    image:
      reference: <registry>/kata-deploy
      tag: "<tag>"

Better long term: bundle the nydus linux-static build in the kata-deploy Dockerfile upstream.

This fixes only Nydus. The kata host binaries (containerd-shim-kata-v2, kata-runtime, kata-monitor) have the same glibc-on-musl problem, but kata-containers ships no static/musl host artifact, so
those must be built against musl or with CGO disabled.

mudler-agent:
Triage: this PR is still relevant to kairos#3932, so I do not think it should be closed as superseded. However, the current diff should not merge as written. It says Hadron works with upstream Confidential Containers on AWS SEV-SNP, while mudler explicitly scoped kairos#3932 to non-Hadron support, and the later testing in this PR found that the stock Kata/Nydus host binaries do not run on musl and that AWS SNP guests do not expose the nested virtualization required by the Kata path.

Please either remove the Hadron/AWS claim and keep this focused on BYOI/non-Hadron installation, or update the text with the documented limitations and link separate Hadron tracking. If nobody plans to revise it, closing this PR and opening a smaller non-Hadron docs change would be clearer.

jimmykarily:
Rebased this PR and updated with the current state of things. We are still waiting for a kata-containers release that will ship the fully static binaries, so that it works with Hadron out of the box. In the meantime, we can merge this PR and complete the non-Hadron part.

PR #607, opened 2026-05-28
Document no-op upgrades

jimmykarily:
Needs: https://github.com/kairos-io/kairos-operator/pull/109/changes


mauromorales (approved):

jimmykarily:
Turns out this solution is not good. See my comment here: https://github.com/kairos-io/kairos-operator/pull/109#issuecomment-4593233540

I will update this PR or open a new one as soon as I have a proper solution.

PR #613, opened 2026-06-08
docs: Clarify partitioning instructions in configuration

Itxaka:
- Updated instructions to specify that setting `fs` to "none" will create a partition without formatting.
- Added note to prevent installation errors when no filesystem is desired for a partition.

### DIFF
--- a/docs/advanced/configuring_partitions.md
+++ b/docs/advanced/configuring_partitions.md
@@ -71,7 +71,8 @@ install:
 
 Note that there are some caveats in the `extra partitions` setup:
  - Only `size`, `fs`, `name` and `label` are used for the partition creation, the name is currently used for the partition label.
- - If a partition has no fs set, the partition will be created, but it will not be formatted.
+ - If `fs` is not set for an extra partition, it defaults to `ext2`.
+ - If a partition should be created without formatting, set `fs` to `none`.
  - No mounting of any type is done during installation to the extra partitions. That part should be done on the stages of the cloud-config manually, with something like the following step:
 
 ```yaml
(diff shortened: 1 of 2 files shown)


docs/advanced/configuring_partitions.md
```diff
@@ -82,6 +82,9 @@ Note that there are some caveats in the `extra partitions` setup:
         - mount -o rw /dev/disk/by-label/PARTITION_TWO /opt/extra
 ```
 
+If not filesystem is wanted for a given partitions, the `fs` field whould be set to `none`, otherwise it will interpreted as an error and the installation will fail.
```
jimmykarily:
```suggestion
If no filesystem is wanted for a given partition, the `fs` field should be set to `none`, otherwise it will interpreted as an error and the installation will fail.
```

So when set to anything other than `none`, it's considered an error and installation fails? This sentence is confusing to me.

docs/reference/configuration.md
```diff
@@ -100,7 +100,7 @@ install:
   # extra partitions to create during install
   # only size, label and fs are used
   # name is used for the partition label, but it's not really used during the kairos lifecycle. No spaces allowed.
-  # if no fs is given the partition will be created but not formatted
+  # if no partitioning is wanted, set the fs to "none" and it will be created but not formatted
```
jimmykarily: "if no partitioning is wanted" means the user doesn't want any partitions to be created. Setting to `none` doesn't prevent the creation of the partition (as the rest of the sentence also explains).

docs/advanced/configuring_partitions.md
Itxaka: This is actually wrong, it defaults to ext2 if no fs is set. Dammit

Itxaka:
@copilot this is wrong, if nothing is mentioned in the fs of the partitions it will default to ext2 as it has always do. Send a PR agains thtis branch to fix the docs

jimmykarily (approved):

PR #615, opened 2026-06-11
Add blog post about core-to-standard using sysext

jimmykarily:
A blog post to present a PoC and gather feedback.

If this gets approved for publication, I need to create the discussion link first and replace it in the blog post before merging. Then I'll put the blog post link in the discussion (so each one references the other).


mauromorales (approved):
This is very nice, I like the idea of trying to get feedback and you highlight very well the trade-offs. So for me it's a go :shipit:

PR #634, opened 2026-07-08
docs(operator): GitOps install, supply-chain security, and NodeOpUpgrade name-bump pattern

ocalzi:
## What this adds

Three interconnected additions covering the full GitOps + supply-chain pipeline for the Kairos operator.

---

### 1. GitOps install via ArgoCD (`installation.md`)

The operator has no Helm chart. This documents deploying it via ArgoCD pointing at the upstream kustomize source, with two non-obvious pitfalls:

- `ServerSideApply=true` is required — the operator CRDs exceed the 262 kB client-side apply annotation limit
- CRD race condition on first bootstrap: downstream `NodeOpUpgrade` / `NodeOp` / `OSArtifact` CRs fail dry-run because CRDs aren't registered yet, even with sync-wave ordering

Two mitigations documented (use both): `SkipDryRunOnMissingResource=true` on downstream CRs + sync-wave ordering:

| Wave | Resource |
|---|---|
| `-20` | `Application/kyverno` (chart + ClusterPolicy resources) |
| `-10` | `Application/kairos-operator` |
| `0` | `NodeOpUpgrade`, `NodeOp`, `OSArtifact` CRs |

---

### 2. Supply-chain security (`supply-chain.md`, new page)

The operator pulls images that run as root on the host and write to the node's A/B partition. Two Kyverno admission gates documented:

**Policy 1 — restrict image source**: `ClusterPolicy` enforcing `NodeOpUpgrade.spec.image` must match `quay.io/kairos/hadron:*`. Blocks misdirected upgrades.

**Policy 2 — verify cosign signature**: Kyverno `verifyImages` with keyless attestor pinned to the Hadron CI identity:
- issuer: `https://token.actions.githubusercontent.com`
- subject: `https://github.com/kairos-io/hadron/.github/workflows/build-multiarch-images.yml@refs/tags/*`
- `mutateDigest: true` prevents tag-swap after check
- Air-gapped note: point `rekor.url` at local Rekor instance

Includes threat model (what is and isn't gated) and bootstrap wave ordering cross-linked with `installation.md`.

---

### 3. NodeOpUpgrade GitOps name-bump pattern + Renovate fix (`nodeop-upgrade.md`)

Documents the GitOps alternative to `generateName`: embed the version in `metadata.name` and bump it alongside `spec.image` in the same commit so the operator sees a new one-shot CR on every merge.

Complete Renovate custom-manager config with **two** `matchStrings` entries — one for `spec.image`, one for `metadata.name`:

- `currentValueTemplate` converts dash-format `v0-3-0` → `v0.3.0` for datasource comparison
- `autoReplaceStringTemplate` writes new version back in dash-format
- Both managers share the same `depNameTemplate` → atomic update in one PR

**Bug documented**: `extractVersionTemplate` is not a valid Renovate custom-manager field and is silently ignored. Using it causes `spec.image` to be bumped but `metadata.name` to stay at the old version — the operator sees no new CR and does nothing.

## Test plan

- [ ] Verify Docusaurus build passes (cross-links between `installation.md` → `supply-chain.md` and `nodeop-upgrade.md`)
- [ ] Verify `supply-chain.md` appears in the operator-docs sidebar
- [ ] Verify cosign policy subject matches current Hadron CI workflow filename (`build-multiarch-images.yml` — confirmed present in repo)

### DIFF
--- a/operator-docs/installation.md
+++ b/operator-docs/installation.md
@@ -48,3 +48,77 @@ To remove the operator installed via bundle, you need to delete the `kairos-oper
 
 - **k0s**: `/var/lib/k0s/manifests/kairos-operator/`
 - **k3s**: `/var/lib/rancher/k3s/server/manifests/`
+
+## Installing via GitOps (ArgoCD)
+
+The operator does not currently publish a Helm chart, so GitOps deployments point directly at the upstream `config/default` kustomize source and let ArgoCD render it. A minimal `Application` looks like:
+
+```yaml
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: kairos-operator
+  namespace: argocd
+  annotations:
+    # Sync-wave places the operator ahead of any NodeOpUpgrade / NodeOp / OSArtifact
+    # CRs applied from the same app-of-apps root.
+    argocd.argoproj.io/sync-wave: "-10"
+spec:
+  project: default
+  source:
+    repoURL: https://github.com/kairos-io/kairos-operator.git
+    targetRevision: v0.0.7                # pin to a released tag; bump via a PR
+    path: config/default
+  destination:
+    server: https://kubernetes.default.svc
+    namespace: kairos-operator
+  syncPolicy:
+    automated:
+      prune: true
+      selfHeal: true
+    syncOptions:
+      - CreateNamespace=true
+      - ServerSideApply=true               # required — the CRDs exceed the 262 kB annotation size limit of client-side apply
+```
+
+Notes:
+
+- `path: config/default` matches the `kubectl apply -k` install above. Both call the same kustomization; ArgoCD just materializes the result.
+- `ServerSideApply=true` is required. The operator's CRDs (in particular `OSArtifact`) exceed the 262 kB `last-applied-configuration` annotation limit and fail with `metadata.annotations: Too long` under client-side apply.
+- `syncOptions.CreateNamespace=true` lets ArgoCD create `spec.destination.namespace` on first sync. The upstream kustomization creates its own `operator-system` namespace; the value above just gives unscoped resources a home.
+
+### CRD race condition when applying downstream CRs
+
+If you commit a `NodeOpUpgrade` / `NodeOp` / `OSArtifact` CR in the same app-of-apps root as the operator install, ArgoCD's pre-sync dry-run validates each resource against the cluster's discovery cache. On the first bootstrap the operator's CRDs are not yet registered, so the dry-run fails with:
+
+```
+the server could not find the requested resource (kind=NodeOpUpgrade)
+one or more synchronization tasks are not valid
+```
+
+Even with `sync-wave` ordering the operator's `Application` reports "Synced" the moment its manifest lands in Kubernetes — not when the CRDs are actually reconciled. The downstream CR then races ahead and 5× retry-fails.
+
+Two mitigations, use both:
+
+**1. Annotate downstream CRs to defer dry-run** until the CRD appears:
+
+```yaml
+apiVersion: operator.kairos.io/v1alpha1
+kind: NodeOpUpgrade
+metadata:
+  name: kairos-upgrade
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+spec:
+  # …
+```
+
+**2. Sequence with sync-waves** — put the operator's `Application` at a lower wave than the CRs so ArgoCD attempts the operator install first. A typical layout:
+
+| Wave | Resource |
+|---|---|
+| `-20` | `Application/kyverno` (Helm chart + `ClusterPolicy` resources) — see [Supply-chain security](supply-chain#bootstrap-ordering-with-argocd) |
+| `-10` | `Application/kairos-operator` |
+| `0` (default) | `NodeOpUpgrade`, `NodeOp`, `OSArtifact` CRs |
+
+`SkipDryRunOnMissingResource=true` is the actual mechanism that unblocks the sync; sync-wave ordering just reduces how many times ArgoCD retries before the CRD shows up.
--- a/operator-docs/installation.md
+++ b/operator-docs/installation.md
@@ -117,7 +117,7 @@ spec:
 
 | Wave | Resource |
 |---|---|
-| `-20` | `Application/kyverno` (Helm chart + `ClusterPolicy` resources) — see [Supply-chain security](supply-chain#bootstrap-ordering-with-argocd) |
+| `-20` | `Application/kyverno` (Helm chart + `ClusterPolicy` resources) — see [Supply-chain security](../supply-chain#bootstrap-ordering-with-argocd) |
 | `-10` | `Application/kairos-operator` |
 | `0` (default) | `NodeOpUpgrade`, `NodeOp`, `OSArtifact` CRs |
 
--- a/operator-docs/nodeop-upgrade.md
+++ b/operator-docs/nodeop-upgrade.md
@@ -86,7 +86,7 @@ Three things to note:
 - The second custom manager handles `metadata.name`. `currentValueTemplate` converts the dash-format slug (`v0-3-0` → `v0.3.0`) so Renovate can compare it against the docker datasource. `autoReplaceStringTemplate` writes the new version back in dash-format. Both managers share the same `depNameTemplate`, so Renovate updates `spec.image` and `metadata.name` atomically in one PR.
 - **Do not use `extractVersionTemplate`** to parse the dash-format in `metadata.name` — it is not a valid field for Renovate custom managers and is silently ignored. The result is that `spec.image` gets bumped but `metadata.name` stays at the old version, so the operator sees no new CR and does nothing.
 
-The pattern is not superior to `generateName` — it's a different tradeoff. Pick `generateName` when you drive upgrades from a CLI or CI job. Pick static-name bump when the desired state lives in git and every change goes through a merged PR.
+`generateName` and static-name bump are equivalent in semantics — both produce a new one-shot CR. The difference is operational: `generateName` suits imperative workflows (CLI, CI job) where the run doesn't need to live in git. Static-name bump suits GitOps where every change is a reviewed commit.
 
 ## Basic Example
 
--- a/operator-docs/nodeop-upgrade.md
+++ b/operator-docs/nodeop-upgrade.md
@@ -82,7 +82,7 @@ For [Renovate](https://docs.renovatebot.com/), a custom regex manager targeting 
 
 Three things to note:
 
-- `allowedVersions` clamps updates to a single k3s minor line. This prevents a Renovate PR from silently proposing a k3s minor bump alongside a Kairos patch bump. To cross a k3s minor, edit the regex in a separate PR — forces an explicit decision.
+- `allowedVersions` pins updates to the current k3s minor line (`1.35.x`). This prevents Renovate from silently bundling a k3s minor upgrade with a Kairos patch bump. To adopt a new k3s minor, update the regex explicitly.
 - The second custom manager handles `metadata.name`. `currentValueTemplate` converts the dash-format slug (`v0-3-0` → `v0.3.0`) so Renovate can compare it against the docker datasource. `autoReplaceStringTemplate` writes the new version back in dash-format. Both managers share the same `depNameTemplate`, so Renovate updates `spec.image` and `metadata.name` atomically in one PR.
 - **Do not use `extractVersionTemplate`** to parse the dash-format in `metadata.name` — it is not a valid field for Renovate custom managers and is silently ignored. The result is that `spec.image` gets bumped but `metadata.name` stays at the old version, so the operator sees no new CR and does nothing.
 
--- a/operator-docs/nodeop-upgrade.md
+++ b/operator-docs/nodeop-upgrade.md
@@ -18,6 +18,76 @@ A `NodeOpUpgrade` represents a **single upgrade run** on the target nodes. The o
 
 To run the same upgrade configuration repeatedly, use **metadata.generateName** instead of **metadata.name** and **kubectl create** (not **apply**) so each run creates a new NodeOpUpgrade. See [NodeOp: One-off operations and reusing manifests](../nodeop/#one-off-operations-and-reusing-manifests) for the same pattern and rationale.
 
+### GitOps alternative: static-name bump with a versioned suffix
+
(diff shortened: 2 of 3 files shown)


Itxaka (commented):
Review notes. Conflicts with #644 — both touch `operator-docs/installation.md`. If #644 lands first, several things here go stale.

**Cross-PR / rebase**

- `operator-docs/installation.md`: "The operator does not currently publish a Helm chart, so GitOps deployments point directly at the upstream `config/default` kustomize source." Directly contradicts #644. On rebase, drop that sentence and add an ArgoCD *Helm source* variant alongside the kustomize path (e.g. `spec.source.chart: kairos-operator` + `repoURL: https://kairos-io.github.io/kairos-operator/charts`). Many GitOps users will prefer the chart route once it exists.

**`operator-docs/nodeop-upgrade.md` — Renovate config**

- Claim: "Both managers share the same `depNameTemplate`, so Renovate updates `spec.image` and `metadata.name` atomically in one PR." Renovate groups by branch name derived from `depName` + `newValue`. Same-branch grouping usually produces one PR in practice, but it isn't a documented atomicity guarantee. Add `groupName` explicitly to lock this in against future Renovate behavior drift.

- `versioningTemplate` regex hardcodes `standard-(?:amd64|arm64)` — the arch fragment becomes part of version comparison. If a user's CR file names use a different arch (or none), Renovate treats them as different packages. Either drop arch from the versioning regex or note this constraint.

- `packageRules.allowedVersions: /.*-k3s-v1\.35\.\d+-k3s1$/` silently drops all updates outside k3s 1.35 — good safety, and the doc mentions bumping it, but consider bolding that operational note.

**`operator-docs/supply-chain.md`**

- **`failurePolicy: Fail` + `webhookTimeoutSeconds: 30` + `background: false`** means during a Sigstore outage, **every** `quay.io/kairos/hadron:*` pod admission blocks — including future operator-driven upgrade pods. The doc names the outage risk but not the "your cluster's upgrade path itself is offline until Sigstore returns" implication. Worth calling out explicitly, plus a suggestion that operators consider `failurePolicy: Ignore` for non-safety-critical clusters or run a local Rekor.

- **Threat model missing an entry**: attacker with Kyverno-admin who edits/deletes the ClusterPolicy trivially bypasses both gates. Add a "policy tampering" bullet and/or a note that Kyverno policy objects themselves need RBAC guards (or GitOps-owned + `preventDeletion` sync options).

- **Policy 1 pattern `image: "quay.io/kairos/hadron:*"`** — Kyverno glob matches one path segment; won't allow digest-pinned form (`quay.io/kairos/hadron@sha256:...`). If digest pinning is desired for reproducibility, switch to a regex validator. If not, note that the policy forces tag references.

- **"Because Kyverno cannot re-evaluate `spec.image` under a `generateName` create (the CR does not yet have a fully-qualified name), match on `CREATE` and `UPDATE` as above."** This reasoning looks inverted — Kyverno's admission review sees `spec.image` regardless of whether `metadata.name` is materialized from `generateName`. `CREATE` matches `generateName` creates fine. Recommend rewording or deleting that clause.

- **Cosign subject**: `.../.github/workflows/build-multiarch-images.yml@refs/tags/*` — please link the actual workflow file inline in the doc so future readers (and reviewers) can spot-check drift when upstream Hadron CI renames anything.

- Bootstrap sync-wave table is nicely consistent between `installation.md` and `supply-chain.md`. Nice.

Overall the pipeline story is great — GitOps + admission gates + Renovate all documented in one arc. Just needs the #644 rebase and the tightenings above.

Itxaka:
note that because I dont know much k8s I used AI to help me review this so take hte comments with a pinch of salt and maybe lets wait for one of our k8s experts to comments on this :D

jimmykarily:
I haven't used Kyverno much, neither ArgoCD (I'm a more of a FluxCD person) but it looks good to me. Good idea documenting this. 2 AIs did a review already so I'll leave you to it.

ocalzi:
I could do an Flux one too. I’ve never touched flux either so could be interesting. I’m aware that this is linked to other PR but I have no problem correcting it as the others PR changes. I will have a proper look at your AI feedback no worries.

ocalzi:
Addressed all review comments + added a Flux section (per @Itxaka's question):

- `supply-chain.md`: both `ClusterPolicy` examples updated `kyverno.io/v2beta1` → `kyverno.io/v1`
- `nodeop-upgrade.md`: aligned the static-name example with the Renovate regex below it — switched from `opensuse` image + generic `kairos-upgrade-v3-4-2` name to `hadron` image + `hadron-mgmt-v0-4-0` name so both `matchStrings` entries match the same CR and Renovate can do the atomic name+image bump as described
- `installation.md`: added a **Flux** section with two options — HelmRelease via OCI (recommended) and Kustomization pointing at `config/default` (same source as the ArgoCD approach above it). CRD race condition covered with `dependsOn` as the Flux equivalent of sync-waves.

Ready for re-review when you get a chance.

jimmykarily (approved):

ocalzi:
Tested the Flux install section on a live Hadron Linux cluster:

- **Hadron**: `v0.4.0` (Kairos `v4.1.2`, kernel `7.1.0-hadron`)
- **k3s**: `v1.35.5+k3s1`
- **Flux**: `v2.9.4` (source-controller `v1.9.4`, helm-controller `v1.6.3`)

Steps validated:
1. `flux install` — all controllers Ready
2. `OCIRepository` (Option A) pointing at an OCI Helm chart → `Ready`
3. `HelmRelease` with `chartRef` → installed successfully

Also caught and fixed a bug in this PR: OCIRepository `apiVersion` was `source.toolkit.fluxcd.io/v1beta2` (removed in Flux v2.3.0) — corrected to `v1` in the latest commit.

mauro-agent (commented):
🤖 **Automated review by `mauro-agent`** — an AI agent operated by @mauromorales. This is not Mauro's personal review; he has not read this PR unless he says so himself.

**Static review only.** I read the diff, the linked repos' workflows, and queried public registries. I did not apply any manifest, run Kyverno, install Flux, or execute `cosign verify`. Where I infer rather than observe, I say so.

## First, a review-coverage gap worth naming

- @jimmykarily's substantive comment (2026-07-22) was *"I haven't used Kyverno much, neither ArgoCD (I'm more of a FluxCD person) … 2 AIs did a review already so I'll leave you to it."*
- The **Flux section did not exist then** — it was added 2026-08-09 in `573f0da2`, eighteen days later.
- @Itxaka's detailed review is also dated 2026-07-22, so it too predates the Flux section.
- The `APPROVED` review landed 2026-08-10 **12:30:24Z**; the head commit `c85d1569` ("correct OCIRepository apiVersion to v1") landed **20:24:18Z** the same day.

So the one section matching the one reviewer's stated expertise arrived after he stepped back, and the approval predates the current head. Not a criticism of anyone — just worth flagging that the Flux content has had no substantive human review, and two of the findings below are in it.

---

## 1. Policy 2's cosign `subject` cannot match any existing signature — copy-pasting this blocks all hadron pods

The policy asserts:

> Upstream Kairos images ([Hadron](https://github.com/kairos-io/hadron)) are cosign-signed at release time by the Hadron CI using a GitHub Actions OIDC identity.

and encodes:

```yaml
subject: "https://github.com/kairos-io/hadron/.github/workflows/build-multiarch-images.yml@refs/tags/*"
```

What I found:

- **`kairos-io/hadron` has no cosign anywhere in CI.** `grep -il cosign` across its entire `.github/` tree returns nothing, including `build-multiarch-images.yml` — the exact file named in the subject.
- **The signatures do exist**, though — sampling `quay.io/kairos/hadron` returns 52 `sha256-….sig` tags out of 100.
- **They are produced by `kairos-io/kairos-factory-action`**, whose `reusable-factory.yaml` runs `cosign sign` under `if: inputs.cosign && inputs.registry_domain != ''`. `kairos-io/kairos`'s `release.yaml` calls it with `cosign: true`.

So the images are signed, but by the factory action invoked from the **kairos** release workflow — not by hadron's CI, which never signs anything. For a reusable workflow GitHub's OIDC `job_workflow_ref` (and hence the Fulcio SAN) is the *called* workflow, so I'd expect the real identity to be something like `https://github.com/kairos-io/kairos-factory-action/.github/workflows/reusable-factory.yaml@<ref>`.

**I did not run `cosign verify`, so I have not observed the actual certificate identity** — that last step is inference. But the direction is not in doubt: it cannot be hadron's `build-multiarch-images.yml`, because that workflow contains no signing step.

Consequence: with `required: true` and `failurePolicy: Fail`, a reader who follows this page verbatim gets **every `quay.io/kairos/hadron:*` pod rejected**, including operator-driven upgrade pods. On a security-hardening page that's the worst failure direction.

The manual command at the bottom has the same problem — `--certificate-identity-regexp='^https://github.com/kairos-io/hadron/'` would also fail.

Suggested fix: run `cosign verify --certificate-oidc-issuer=https://token.actions.githubusercontent.com --certificate-identity-regexp='.*' quay.io/kairos/hadron:<tag>` against a real tag, read the actual `Subject`, and put that value in both the policy and the CLI example. Worth a sentence explaining that signing happens in the factory action, since that is genuinely surprising.

## 2. Flux Option A points at a Helm chart that does not appear to be published

```yaml
url: oci://ghcr.io/kairos-io/helm-charts/kairos-operator
ref:
  tag: "0.1.0"
```

Everything I can check says this chart does not exist:

- `kairos-io/kairos-operator` has **no chart directory** — its top level is `api cmd config examples hack internal test` and friends, no `chart/`, `charts/` or `helm/`.
- **None of its five workflows** (`lint.yml`, `release-orchestrator.yml`, `release.yml`, `test-e2e.yml`, `test.yml`) mentions helm or chart, so nothing publishes one.
- `kairos-io/helm-charts` **redirects to `kairos-community/helm-charts`** — a different org. That repo does publish charts (`index.yaml` → HTTP 200) but its chart list is `cluster-api-provider-kairos, entangle-proxy, entangle, kairos-crds, kcrypt-challenger, osbuilder` — **no `kairos-operator`**.
- The Helm-repo URL that #644 documents, `https://kairos-io.github.io/kairos-operator/charts/index.yaml`, returns **HTTP 404**, and `kairos-operator` has no `gh-pages` branch.
- Anonymous GHCR token requests for `kairos-io/helm-charts/kairos-operator` are refused outright.

I note the test report says *"OCIRepository (Option A) pointing at **an** OCI Helm chart → Ready"* — if that was a locally-published chart rather than this URL, Option A is still unverified. Could you confirm the exact `url`/`tag` that was tested?

This is not only this PR's problem: **#644 documents the same two unpublished locations**, so whichever lands first ships a broken install path. Since Option A is labelled *(recommended)*, it is the one most readers will try.

## 3. Missing required `interval` in the `dependsOn` example

```yaml
kind: Kustomization
metadata:
  name: kairos-upgrades
spec:
  dependsOn:
    - name: kairos-operator
  path: ./upgrades
```

`spec.interval` is required on a Flux `Kustomization` — the two examples above this one both set `interval: 12h`, this one omits it. As written the manifest is rejected at apply time. Same page, so easy to miss by pattern-matching against the neighbours.

## 4. The threat model claims something Policy 1 does not actually deliver

> **Gated:** Attacker with cluster-admin who tries to point `NodeOpUpgrade.spec.image` at their own registry — blocked by Policy 1.

A cluster-admin can `kubectl delete clusterpolicy restrict-nodeopupgrade-image`, or edit the Kyverno deployment, and then do as they like. Policy 1 gates a *`NodeOpUpgrade` author*, not a cluster-admin. As written the page overstates its own guarantees in the one section readers will use to reason about risk.

This is @Itxaka's "policy tampering" point, which I don't think was picked up — there's still no such bullet under **Not gated**. Two edits: downgrade the cluster-admin claim to something like "an attacker who can create `NodeOpUpgrade` CRs but cannot modify cluster policy", and add the tampering entry with its mitigation (RBAC on `ClusterPolicy`, or GitOps-owned policies with deletion protection).

## 5. Other items from @Itxaka's review that look unaddressed

- **`groupName`** — line still reads *"Both managers share the same `depNameTemplate`, so Renovate updates `spec.image` and `metadata.name` atomically in one PR."* No `groupName` was added to the JSON. His point stands: same-branch grouping is emergent behaviour, not a documented guarantee.
- **Arch fragment in `versioningTemplate`** — `standard-(?:amd64|arm64)` is still baked into both managers with no note about the constraint it imposes on users whose CR filenames differ.
- **The inverted Kyverno/`generateName` clause** — unchanged, verbatim: *"Because Kyverno cannot re-evaluate `spec.image` under a `generateName` create (the CR does not yet have a fully-qualified name), match on `CREATE` and `UPDATE` as above."* I agree with his read: the AdmissionReview carries `spec.image` whether or not `metadata.name` is materialised, so the stated reason is not the real one. Matching `CREATE` + `UPDATE` is right; the justification is wrong and should be reworded or dropped.
- **Link the cosign workflow file inline** — currently links the workflows *directory*, not the file. Given finding 1, this matters more than it looked: an inline link would probably have surfaced that hadron doesn't sign.
- **`failurePolicy: Fail` blast radius** — half covered. The doc now says a Sigstore outage blocks hadron pod admission, but not his sharper point: that this takes your *upgrade path itself* offline, and that `failurePolicy: Ignore` may be the right trade for non-safety-critical clusters.

## 6. Smaller things

- **Policy 1's glob rejects digest-pinned images.** `image: "quay.io/kairos/hadron:*"` matches tags only, so `quay.io/kairos/hadron@sha256:…` is refused — while Policy 2's `mutateDigest: true` is teaching readers that digests are the safer reference. Worth a note, or a regex validator.
- **Pinned versions are already stale.** Both Flux options pin `0.1.0` / `v0.1.0`; `kairos-operator`'s latest release is **v0.1.3** (2026-08-06).
- **#644 is still open** and still touches `operator-docs/installation.md`, so @Itxaka's conflict warning remains live.

## What's clearly good

- The `kyverno.io/v2beta1` → `v1` fix landed on both policies.
- Testing the Flux path on a real Hadron/k3s cluster, reporting versions, and catching the `source.toolkit.fluxcd.io/v1beta2` removal is exactly the kind of verification docs like this usually lack.
- `mutateDigest: true` with the tag-swap rationale is a genuinely good detail.
- The sync-wave tables agree between `installation.md` and `supply-chain.md`.
- The overall arc — GitOps install, admission gates, Renovate-driven upgrades — is a story the docs didn't tell before and should.

Findings 1 and 2 are the ones I'd want resolved before merge; both are cases where following the page produces a broken or blocked cluster.

mudler-agent:
Triage update: #644 merged into `main` on August 19, so this branch now needs a rebase and reconciliation of `operator-docs/installation.md`. The overall GitOps and supply-chain documentation is still relevant, so this PR is not superseded as a whole. However, it should not merge in its current state: the unresolved review findings include a cosign identity that does not match the signing workflow and a recommended Flux Helm source that was not published when checked. Please update and re-verify those examples against current `main`; if that work is no longer planned, this PR should be closed rather than left open with copy-pasteable broken installation/security guidance.

PR #636, opened 2026-07-09
docs: explain how to extract the raw image from the -img artifact

### LINKED ISSUE #4232: Raspberry Pi docs don't explain how to obtain/extract the raw image
## What's missing

The [Raspberry Pi installation page](https://kairos.io/docs/installation/raspberry/) doesn't explain how to obtain the actual raw image to flash. For arm64/RPi, raw artifacts are not attached to GitHub releases (an RPi4 raw is ~5 GB, over GitHub's 2 GiB per-file release-asset limit), so users need another route — but the page describes none. Anyone following the docs to build/download an RPi image reaches a dead end.

## Suggested addition

The build pipeline publishes the raw as an OCI artifact at `<registry>/<repo>:<tag>-img`, built from [`images/Dockerfile.img`](https://github.com/kairos-io/kairos/blob/master/images/Dockerfile.img) (`FROM scratch` + `COPY $ARTIFACT /$ARTIFACT`). It can be extracted with plain Docker:

```bash
export IMAGE=quay.io/kairos/<flavor>:<tag>-img
container=$(docker create "$IMAGE" noop)   # scratch image: throwaway cmd, never executed
docker cp "$container:/artifacts/." .      # the raw is copied to /artifacts
docker rm "$container"
# the .raw image is now in the current directory
```

(Verified against a `kairos-factory-action@v1.1.3` raw build — this pulls the full ~5 GB `.raw` out of the `-img` artifact.)

Could the Raspberry Pi page get an "extract the raw image" section along these lines (or an equivalent `crane export` / `skopeo` approach)?

## Related

Companion to #4231, which proposes having the factory action publish a release with these instructions automatically.

mauromorales:
## What

The Raspberry Pi installation page tells users to "extract the img file from a container image as described in [the image matrix page]", but that reference page no longer documents *how* to do it — the previous `luet`-based instructions were removed, leaving a dead end for anyone trying to obtain a raw/RPi image.

This adds an **"Extracting a raw image from a container image"** section to the image matrix reference, explaining the Docker-only method (the `-img` tag is a `scratch` image that carries the raw under `/artifacts`), and links the Raspberry Pi "Download" step directly to it.

## Why Docker (no luet)

The `-img` artifact is built from `images/Dockerfile.img` (`FROM scratch` + `COPY $ARTIFACT /$ARTIFACT`), so `docker create` + `docker cp` pulls the raw out with no extra tooling. Verified end to end against a `kairos-factory-action@v1.1.3` RPi4 build — it extracts the full ~5 GB `.raw`.

Fixes kairos-io/kairos#4232 — companion to kairos-io/kairos#4231 (automating the same instructions in the factory action's release notes).

### DIFF
--- a/docs/installation/edge-devices/raspberry.md
+++ b/docs/installation/edge-devices/raspberry.md
@@ -48,7 +48,7 @@ sudo dd if=build/{{< Image variant="standard" model="rpi4" arch="arm64" suffix="
 
 ### Download
 
-Extract the `img` file from a container image as described [in this page](/docs/reference/image_matrix)
+Extract the raw image from the published `-img` container image as described in [Extracting a raw image from a container image](/docs/reference/image_matrix#extracting-a-raw-image-from-a-container-image).
 
 ### Flash the image
 
--- a/docs/reference/image_matrix.md
+++ b/docs/reference/image_matrix.md
@@ -55,6 +55,31 @@ docker build --platform linux/arm64 --build-arg BASE_IMG=almalinux:9 --build-arg
 :::tip Note
 See [Kairos Factory](/docs/reference/kairos-factory/) for a production-ready path to automate this in CI.
 :::
+
+### Extracting a raw image from a container image
+
+Raw disk images (for example the Raspberry Pi `.raw`) are not attached to GitHub releases, because they exceed GitHub's 2 GiB per-file asset limit. Instead they are published as an OCI artifact whose tag ends in `-img`.
+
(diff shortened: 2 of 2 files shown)


Itxaka (approved):

Itxaka:
5Gb may be the old ones though...I though the newer are smaller due to hadron + recovery shipping only trick. See, before we shipped the full image, so active+passive+recovery but now with only recovery images should not be bigger than a Gb tops I think??

Itxaka (approved):

mauromorales:
@Itxaka good point, should we try to also publish an artifact then? instead of having those on quay?

PR #644, opened 2026-07-19
docs(operator): add Helm and Kustomize install.yaml installation methods

ocalzi:
## Summary

- Adds Helm as the recommended installation method (Helm repo + OCI registry)
- Adds key values table covering `leaderElect`, `toolImage`, `buildahImage`, `sentinelImage`, `nodeops.defaultImage`, `tolerations`
- Adds CRD upgrade warning (`crds/` dir — Helm never upgrades/deletes them)
- Adds pre-built `install.yaml` from GitHub Release assets as the primary Kustomize path (no `git` required); keeps `kubectl apply -k` as alternative
- Removes stale statement that no Helm chart exists
- Restructures page with clear sections and removal instructions per method

## Related

Depends on / tracks: kairos-io/kairos-operator#137

### DIFF
--- a/operator-docs/installation.md
+++ b/operator-docs/installation.md
@@ -6,29 +6,109 @@ date: 2025-07-25
 description: Deploy and remove the Kairos operator on your Kubernetes cluster
 ---
 
-## Deploying the operator
+## Helm (recommended)
 
-To deploy the operator, you can use kubectl (provided that the `git` command is available):
+The operator publishes a Helm chart on each release via GitHub Pages and the GHCR OCI registry.
+
+### From the Helm repository
 
 ```bash
-# Using GitHub URL
-kubectl apply -k https://github.com/kairos-io/kairos-operator/config/default
+helm repo add kairos-operator https://kairos-io.github.io/kairos-operator/charts
+helm repo update
+helm install kairos-operator kairos-operator/kairos-operator \
+  --namespace kairos-operator --create-namespace
+```
 
-# Or using local directory (if you have the operator checked out)
-kubectl apply -k config/default
+### From the OCI registry
+
+```bash
+helm install kairos-operator \
+  oci://ghcr.io/kairos-io/helm-charts/kairos-operator \
+  --version 0.1.0 \
+  --namespace kairos-operator --create-namespace
+```
+
+### Key values
+
+| Value | Default | Description |
+|-------|---------|-------------|
+| `image.operator.tag` | Chart `appVersion` | Operator image tag |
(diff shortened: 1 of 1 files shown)


Itxaka (approved):
sounds good to me, thanks!

Itxaka (commented):
Review notes. Conflicts with #634 — both edit `operator-docs/installation.md`. Whichever lands second needs rebase.

**Findings**

- **CRD upgrade note applies whole bundle, not CRDs only.** The `kubectl apply -f .../releases/latest/download/install.yaml` snippet under "CRDs are placed in the `crds/` directory" upgrades the entire bundle (Deployment, RBAC, ServiceAccount, CRDs). Run against a Helm-managed release it will fight Helm's ownership of the Deployment. Options: publish a CRDs-only release asset, or pipe through a filter like `kubectl apply --server-side -f <(curl -sL .../install.yaml | yq 'select(.kind == "CustomResourceDefinition")')`. At minimum, add a warning that this snippet is not safe against a Helm-managed install.

- **Hardcoded `--version 0.1.0` in the OCI example** will pin readers to a stale version once new releases ship. Use a placeholder (`--version <VERSION>`) or reference the [releases page](https://github.com/kairos-io/kairos-operator/releases). Same treatment as the Helm repo path would be nice too — add optional `--version` there for parity.

- **Key values table needs a spot-check against upstream `values.yaml`.** Verify `image.operator.tag`, `image.nodeLabeler.tag`, `sentinelImage`, `nodeops.defaultImage`, `toolImage`, `buildahImage` all map to actual chart values with the described defaults.

- **`nodeops.defaultImage` default `busybox:latest`** — documenting `:latest` as a default. Product-side decision, but flag if unintended for reproducibility.

- Nit: consider adding a `--version` example (or note about picking a version) to the Helm repo install path for consistency with the OCI section.

Otherwise LGTM — nice restructure and the Helm-first ordering reads well.

Itxaka:
note that because I dont know much k8s I used AI to help me review this so take hte comments with a pinch of salt and maybe lets wait for one of our k8s experts to comments on this :D

ocalzi:
Addressed the review comment in the latest commit:

- `leaderElect` example: `true` → `false`. The default is already `true` (as noted in the values table on the same page), so showing `--set leaderElect=true` as an override example is a no-op. Switched to `false` to demonstrate the actual useful override: disabling leader election for single-replica dev installs.

The `install.yaml` asset comments are now resolved by the operator-side change in kairos-operator#137, which adds `gh release upload install.yaml` to the release workflow — no doc change needed here.

Ready for re-review when you get a chance.

mauromorales (approved):

PR #659, opened 2026-08-06
docs(operator): resource limits/requests for build pods

thebhdn:
Hey folks,

Docs for https://github.com/kairos-io/kairos-operator/issues/41

### DIFF
--- a/operator-docs/osartifact.md
+++ b/operator-docs/osartifact.md
@@ -1054,6 +1054,82 @@ spec:
 
 ---
 
+## Resource specification
+
+By default the builder Pod and its containers run with no resource requests or limits. You can use `spec.resources` to set resource constraints per artifact kind and at the pod level. This is useful for controlling resource consumption on shared clusters or ensuring builds have sufficient CPU and memory.
+
+The `spec.resources` field accepts the following optional sub-fields:
+
+| Field | Applies to | Description |
+| ------- | ------------ | ------------- |
+| `iso` | ISO builder container | CPU/memory limits for the ISO build step |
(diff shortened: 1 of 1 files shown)


operator-docs/osartifact.md
```diff
@@ -1054,6 +1054,82 @@ spec:
...
+      requests:
+        memory: "4Gi"
+        cpu: "1"
+      limits:
+        memory: "8Gi"
+        cpu: "4"
+```
+
+:::caution Pod-level resources
+The **PodLevelResources** feature gate must be enabled on your cluster (beta since Kubernetes 1.32). If the gate is not present, the `pod` field is silently ignored by the API server. When both `pod` and per-kind resources are set, Kubernetes enforces that pod-level limits are at least the sum of all container-level limits.
```
jimmykarily: A little clearer on what makes the Pod unschedulable:

```suggestion
The PodLevelResources feature gate must be enabled on your cluster. It went alpha in Kubernetes 1.32 and beta (on by default) in 1.34. If the gate is disabled, the pod field is silently dropped by the API server.
When both pod and per-kind resources are set, Kubernetes will reject the pod at creation if either of these is true:
  - the sum of container requests is greater than the pod's requests, or
  - any single container's limit is greater than the pod's limit.

The sum of container limits is allowed to exceed the pod's limit. That will not cause a rejection; the pod's limit still caps total usage at runtime.
```

jimmykarily (approved):

jimmykarily (approved):

jimmykarily:
Good job!

PR #661, opened 2026-08-07
docs(operator): nameOverride for artifacts

thebhdn:
Hey folks,

docs for [feat: ability to specify generated ISO full name](https://github.com/kairos-io/kairos-operator/issues/85)

### DIFF
--- a/operator-docs/osartifact.md
+++ b/operator-docs/osartifact.md
@@ -481,9 +481,14 @@ By default, all artifacts use the OSArtifact resource's `metadata.name` as the b
 | `spec.nameOverride.azureImage` | string | Override the Azure VHD filename (without `.vhd` extension). |
 | `spec.nameOverride.gceImage` | string | Override the GCE archive filename (without `.gce.tar.gz` extension). |
 | `spec.nameOverride.netboot` | string | Override the netboot artifact filename. |
-| `spec.nameOverride.uki` | string | Override the UKI signed artifact filename (without extension). |
+| `spec.nameOverride.uki` | string | Override the UKI signed artifact base filename. The actual output appends a `-uki` suffix (e.g. `<nameOverride.uki>-uki.iso`). |
 
-Each name must be 1–253 characters long and can contain only lowercase letters, numbers, and hyphens `-`. It must start and end with a letter or number, and must not contain dots `.` or slashes `/`. If a specific name is not set, the artifact uses the resource's `metadata.name`
+Each name must be 1–253 characters long and can contain only lowercase letters, numbers, and hyphens `-`. It must start and end with a letter or number. If a specific name is not set, the artifact uses the resource's `metadata.name`.
+
+
+:::warning Warning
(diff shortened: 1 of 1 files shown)


jimmykarily:
**[Claude review]** (drafted by Claude, posted on my behalf)

Reviewed the doc changes against the operator implementation on `main` (feature added in kairos-operator commit 8338731 / PR kairos-io/kairos-operator#154). The field list, extensions, and validation pattern (`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`, 1 to 253 chars) all match `api/v1alpha2/osartifact_types.go`. A few things worth addressing:

### 1. Example won't validate: `baseImage` is missing

The new example sets `image.buildOptions.version` but omits `baseImage`. `OSArtifactSpec.Validate()` in `api/v1alpha2/osartifact_types.go` rejects this:

```go
if img.BuildOptions.BaseImage == "" {
    return fmt.Errorf("spec.image.buildOptions.baseImage is required when using buildOptions")
}
```

Every other `buildOptions` example in this same file includes `baseImage`. Suggest adding e.g. `baseImage: ubuntu:24.04`.

### 2. UKI section understates the filename

The doc says: "Override the UKI signed artifact filename (without extension)."

But `ukiArtifactName` in `internal/controller/job.go` always appends `-uki`:

```go
// ...so they do not collide with unsigned artifacts (e.g. <name>-uki.iso vs <name>.iso)
func ukiArtifactName(artifactName string) string { return artifactName + "-uki" }
```

So `nameOverride.uki: prod` actually produces `prod-uki.iso`, `prod-uki.efi`, etc., not `prod.iso`. Worth calling out the `-uki` suffix explicitly so users are not surprised.

### 3. Netboot coupling is invisible (real footgun)

`isoBasenameForNetboot` in `internal/controller/job.go` uses `NameOverride.Netboot` (falling back to `metadata.name`) to locate the source ISO fed into `auroraboot netboot`. But the ISO on disk was written using `NameOverride.ISO`. So if a user sets:

```yaml
nameOverride:
  iso: foo
artifacts:
  iso: true
  netboot: true
```

the ISO lands at `/artifacts/foo.iso` while the netboot step looks for `/artifacts/<metadata.name>.iso` and the build fails.

The doc currently implies each override is independent. Options:
- Add a caveat: "when netboot is enabled, `nameOverride.netboot` must equal `nameOverride.iso`."
- Better: fix in the operator so netboot reuses the ISO name (open a follow-up issue on kairos-operator?). The coupling is invisible from the CRD schema, so I would lean toward fixing it upstream.

### 4. Minor nits

- Missing trailing period on: "If a specific name is not set, the artifact uses the resource's `metadata.name`"
- The clause "must not contain dots `.` or slashes `/`" is redundant given the preceding "can contain only lowercase letters, numbers, and hyphens". Consider trimming.

Overall the change is useful and well scoped. Item 1 blocks copy-paste; items 2 and 3 are accuracy gaps.

jimmykarily:
The third finding above if a real bug and needs fixing on the operator side. We can fix the rest and make a note for the bug until we fix it.

thebhdn:
> The third finding above if a real bug and needs fixing on the operator side. We can fix the rest and make a note for the bug until we fix it.

Added warning for this specific case. I'll take a look for that bug discovered by Mr. Claude if you don't mind 😄

jimmykarily (approved):

